PR #2, opened 2025-01-30
KMS-510: cdk infrastructure code for deploying rdf4j

cgokey:
The purpose of this code is to use CDK to deploy infrastructure code for deploying the RDF4J database.   It deploys 3 stacks:

rdf4jIamStack - This stack sets up the necessary IAM permissions for ECS tasks to run and be managed, interaction with EFS for file storage, using the ECS Execute Command feature for debugging, and pulling images from ECR.  The exported role ARN can be used by other stacks or services that need to reference this IAM role.

rdf4jEfsStack - This stack essentially sets up a persistent file storage system (EFS) that can be used by ECS tasks.   This is used for storing RDF4j data.  The security group configurations ensure that only the specified ECS tasks can access the EFS. The exported IDs can be used by other stacks or services that need to reference this EFS setup.

rdf4jEcsStack - This stack essentially sets up a containerized RDF4J service running on ECS Fargate, with persistent storage provided by EFS, and accessed through an Application Load Balancer. It's designed to run a single instance of the service, with auto-scaling configured to maintain exactly one task. The service  integrates with other AWS services like ECR for container images and CloudWatch for logging.  One thing to note. Multiple processes can read from the RDF db, but only 1 can write.   A future ticket will setup a lock manager so we can scale, but for now, we will prevent any scaling.

Finally, this includes a custom configuration for the container that runs in ECS.   It creates a customized RDF4J database environment, tailored for use with AWS EFS for persistent storage. It includes custom configuration and setup scripts, and prepares the environment for running as a non-root user (tomcat) for security purposes.

### DIFF
--- a/infrastructure/rdfdb/cdk/bin/deploy_to_ecr.sh
+++ b/infrastructure/rdfdb/cdk/bin/deploy_to_ecr.sh
@@ -2,8 +2,8 @@
 
 # Set variables
 REPO_NAME="rdf4j"
-REGION="us-east-1"  # Replace with your desired region
-DOCKER_FILE_PATH="../../docker"  # Adjust this path as needed
+REGION="us-east-1"
+DOCKER_FILE_PATH="../../docker"
 
 # Create ECR repository
 aws ecr create-repository --repository-name $REPO_NAME --region "us-east-1"
--- a/infrastructure/rdfdb/cdk/bin/deploy_to_ecr.sh
+++ b/infrastructure/rdfdb/cdk/bin/deploy_to_ecr.sh
@@ -9,7 +9,7 @@ DOCKER_FILE_PATH="../../docker"  # Adjust this path as needed
 aws ecr create-repository --repository-name $REPO_NAME --region "us-east-1"
 
 # Get the repository URI
-REPO_URI=$(aws ecr describe-repositories --repository-names $REPO_NAME --region "us-east-1" --query 'repositories[0].repositoryUri' --output text)
+REPO_URI=$(aws ecr describe-repositories --repository-names $REPO_NAME --region $REGION --query 'repositories[0].repositoryUri' --output text)
 
 # Authenticate Docker to ECR
 aws ecr get-login-password --region $REGION | docker login --username AWS --password-stdin $REPO_URI
--- a/infrastructure/rdfdb/cdk/bin/deploy_to_ecr.sh
+++ b/infrastructure/rdfdb/cdk/bin/deploy_to_ecr.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+# Set variables
+REPO_NAME="rdf4j"
+REGION="us-east-1"  # Replace with your desired region
+DOCKER_FILE_PATH="../../docker"  # Adjust this path as needed
+
+# Create ECR repository
+aws ecr create-repository --repository-name $REPO_NAME --region "us-east-1"
+
+# Get the repository URI
+REPO_URI=$(aws ecr describe-repositories --repository-names $REPO_NAME --region "us-east-1" --query 'repositories[0].repositoryUri' --output text)
+
+# Authenticate Docker to ECR
+aws ecr get-login-password --region $REGION | docker login --username AWS --password-stdin $REPO_URI
+
+# Build Docker image
+docker build -t $REPO_NAME $DOCKER_FILE_PATH
+
+# Tag the image
+docker tag $REPO_NAME:latest $REPO_URI:latest
+
+# Push the image to ECR
+docker push $REPO_URI:latest
+
+echo "Docker image pushed to $REPO_URI:latest"
--- a/infrastructure/rdfdb/cdk/bin/rdf4j-aws-deployment.js
+++ b/infrastructure/rdfdb/cdk/bin/rdf4j-aws-deployment.js
@@ -12,7 +12,7 @@ async function main() {
     region: process.env.CDK_DEFAULT_REGION
   }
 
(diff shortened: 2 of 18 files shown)


infrastructure/rdfdb/cdk/bin/deploy_to_ecr.sh
```diff
@@ -0,0 +1,26 @@
...
+# Set variables
+REPO_NAME="rdf4j"
+REGION="us-east-1"  # Replace with your desired region
+DOCKER_FILE_PATH="../../docker"  # Adjust this path as needed
+
+# Create ECR repository
+aws ecr create-repository --repository-name $REPO_NAME --region "us-east-1"
+
+# Get the repository URI
+REPO_URI=$(aws ecr describe-repositories --repository-names $REPO_NAME --region "us-east-1" --query 'repositories[0].repositoryUri' --output text)
```
eereiter, line 12: You have REGION defined above replace your hard coded version of it with the REGION variable.

infrastructure/rdfdb/cdk/bin/rdf4j-aws-deployment.js
```diff
@@ -0,0 +1,44 @@
...
+
+async function main() {
+  const app = new cdk.App()
+
+  const env = {
+    account: process.env.CDK_DEFAULT_ACCOUNT,
+    region: process.env.CDK_DEFAULT_REGION
+  }
+
+  const vpcId = 'vpc-05fdc01976f74a201' // SIT
```
eereiter, line 15: Shouldn't this be an environment variable?

infrastructure/rdfdb/cdk/lib/efs-stack.js
```diff
@@ -0,0 +1,85 @@
...
+      path: '/rdf4j-data',
+      createAcl: {
+        ownerUid: '101',
+        ownerGid: '65534',
+        permissions: '755'
+      },
+      posixUser: {
+        uid: '101',
+        gid: '65534'
+      }
```
eereiter, line 52: Are all of these hard coded values the same for all of our environments?  If not then these should be variables, otherwise I have no comment.

infrastructure/rdfdb/docker/scripts/setup.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -e
+
+echo 'org.apache.tomcat.util.digester.PROPERTY_SOURCE=org.apache.tomcat.util.digester.EnvironmentPropertySource' >> /usr/local/tomcat/conf/catalina.properties
+echo "RDF4J_DATA_DIR is set to: ${RDF4J_DATA_DIR}"
+
+/usr/local/tomcat/modify_web_xml.sh
```
eereiter, line 7: I am curious why tomcat was chosen.  Is it because that was used in the past?  I don't know if it is better then jetty or others - As long as it is performant I don't have an opinion one way or the other.

eereiter (approved):
Just a few minor changes.

infrastructure/rdfdb/docker/scripts/setup.sh
cgokey: Good question... The original commit had a .gitignore on the Dockerfile, so that is why it wasn't apparent.   I added the Dockerfile as well to this PR.   The Dockerfile using a standard image for RDF4J which runs the database in Tomcat.   We could change it later, if need be to Jetty though and create a custom image.

infrastructure/rdfdb/cdk/lib/efs-stack.js
cgokey: Yeah these are the same from env to env, they are values for inside the running docker container.

infrastructure/rdfdb/docker/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+RUN chmod -R 755 /rdf4j-data
+
+# Set the RDF4J data directory
+ENV RDF4J_DATA_DIR=/rdf4j-data
+
+# Install necessary tools
+RUN apt-get update && apt-get install -y unzip zip
+
+# Copy configuration files
+COPY config/tomcat-users.xml /usr/local/tomcat/conf/tomcat-users.xml
```
jceaser, line 16: tomcat and not jetty?

infrastructure/rdfdb/docker/scripts/modify_web_xml.sh
```diff
@@ -0,0 +1,51 @@
...
+            <web-resource-name>RDF4J Server</web-resource-name>\
+            <url-pattern>/*</url-pattern>\
+        </web-resource-collection>\
+        <auth-constraint>\
+            <role-name>rdf4j-user</role-name>\
+        </auth-constraint>\
+    </security-constraint>\
+\
+    <login-config>\
+        <auth-method>BASIC</auth-method>\
```
jceaser, line 35: can we not use Digest?

infrastructure/rdfdb/docker/scripts/setup.sh
jceaser: I was asking the same question. I think Jetty is cleaner but having it built in makes this simpler.

infrastructure/rdfdb/docker/scripts/modify_web_xml.sh
cgokey: The DB is internal, has no access to the public, so I think we are ok with basic.   I could create a 1 point ticket to switch it though, doesn't look that difficult.

jceaser (approved):

PR #19, opened 2025-02-28
KMS-521: Support legacy JSON

mandyparson:
# Overview

### What is the feature?

The API will also need to be able to retrieve concepts in JSON

JSON, I believe are internal versions, but they maybe used by third parties since these have not been locked down.

### What is the Solution?

Created a utility called toLegacyJson that takes in skosConcept and outputs legacy JSON

utils

# Testing

### Reproduction steps

- **Environment for testing: local
- **Collection to test with: 

1. Import toLegacyJSON.js to getConcept/handler.js and then past in these lines at line 99:
    const concepts = [
      { conceptIRI },
      ...(concept['skos:broader'] ? [{ conceptIRI: concept['skos:broader']['@rdf:resource'] }] : []),
      ...(concept['skos:narrower'] || []).map((narrow) => ({ conceptIRI: narrow['@rdf:resource'] })),
      ...(concept['skos:related'] || []).map((related) => ({ conceptIRI: related['@rdf:resource'] }))
    ]

    const conceptSchemeMap = await createConceptSchemeMap()
    const prefLabelMap = await createPrefLabelMap(concepts)

    const JSONOutput = await toLegacyJSON(concept, conceptSchemeMap, prefLabelMap)
    console.log('🚀 ~ getConcept ~ JSONOutput:', JSONOutput)

3. While running npm run offline, use the curl command to see the different JSON outputs $ curl 'http://localhost:4001/dev/concept/599af9b7-4517-45c2-8505-a46fc204e524'
4. Try out different UUIDs from the convertedConcepts.rdf to ensure that all scenarios look like their legacy json counterpart here: https://gcmd.earthdata.nasa.gov/kms/concept/599af9b7-4517-45c2-8505-a46fc204e524?format=json

A few things have been purposely omitted:
broader { id }
narrower { id }

### Attachments

# Checklist


serverless/src/utils/toLegacyJSON.js
```diff
@@ -0,0 +1,148 @@
+import { XMLParser } from 'fast-xml-parser'
+
+import getConceptSchemes from '@/getConceptSchemes/handler'
+import { getSkosConcept } from '@/shared/getSkosConcept'
+
+// Helper function to retrieve parsed information in concept_schemes xml
+function findMatchingLongName(schemeShortName, parsedData) {
```
cgokey, line 7: make these arrow functions, i.e., const findMatchingLongName = (schemeShortName, parsedData) => {}

serverless/src/utils/toLegacyJSON.js
```diff
@@ -0,0 +1,148 @@
...
+      processedLabel.category = label['@gcmd:category']
+    }
+
+    processedLabel.text = label['@gcmd:text']
+
+    return processedLabel
+  })
+}
+
+const toLegacyJSON = async (conceptIRI, skosConcept) => {
```
cgokey, line 38: Overall this looks really good.  I think the only thing that could be improved is the and we'll need in order to make this fast enough is make toLegacyJson not have to do any async calls, that meams there can't be any await calls in it.   All the information it needs shoudl be passed into the function.

I think to solve this, you just need to pass in 2 maps to this the toLegacyJSON function:
conceptSchemeMap[schemeShortName] and returns the long name.
conceptPrefLabelMap[uuid] and returns the prefLabel.

Then you can use these maps instead of calling await to get this information.  From what I can tell this should work, with the exception of isLeaf for broader, narrower, and in this case, we lets just not include this.

serverless/src/utils/toLegacyJSON.js
```diff
@@ -0,0 +1,148 @@
+import { XMLParser } from 'fast-xml-parser'
```
cgokey, line 1: can this be moved out of util and into the shared folder with the others?

serverless/src/shared/createConceptSchemeMap.js
```diff
@@ -0,0 +1,52 @@
...
+ *   // Concept Scheme Map: Map(2) {
+ *   //   'scienceKeywords' => 'Science Keywords',
+ *   //   'platforms' => 'Platforms'
+ *   // }
+ * } catch (error) {
+ *   console.error('Error creating concept scheme map:', error);
+ * }
+ */
+
+function createConceptSchemeMap() {
```
htranho, line 33: Can this be arrow function like other in shared?

serverless/src/getConcepts/handler.js
```diff
@@ -88,49 +84,110 @@ export const getConcepts = async (event) => {
...
+    switch (format.toLowerCase()) {
+      case 'json': {
+        const jsonResponse = {
+          hits: totalConcepts,
+          page_num: pageNum,
+          page_size: pageSize,
+          termsOfUse: 'https://cdn.earthdata.nasa.gov/conduit/upload/5182/KeywordsCommunityGuide_Baseline_v1_SIGNED_FINAL.pdf',
+          keywordVersion: '20.8',
+          viewer: 'https://gcmd.earthdata.nasa.gov/KeywordViewer/scheme/all',
+          concepts: conceptURIs.map((uri) => {
```
cgokey, line 108: I was thinking this would just be `concepts: skosConcepts.map((concept) => toLegacyJson(concept))`

serverless/src/getConcept/handler.js
```diff
@@ -68,38 +71,67 @@ export const getConcept = async (event) => {
...
     })
+
     const conceptIRI = `https://gcmd.earthdata.nasa.gov/kms/concept/${concept['@rdf:about']}`
-    const rdfJson = {
-      'rdf:RDF': {
-        ...namespaces,
-        'gcmd:gcmd': await getGcmdMetadata({ conceptIRI }),
-        'skos:Concept': [concept]
 
+    // Prepare data for different formats
```
cgokey, line 83: line 83-92 should be moved inside the case 'json' since it is only needed when the output is json.

serverless/src/getConcepts/handler.js
```diff
@@ -88,49 +84,110 @@ export const getConcepts = async (event) => {
...
-      'rdf:RDF': {
-        ...namespaces,
-        'gcmd:gcmd': await getGcmdMetadata({
-          pageNum,
-          pageSize,
-          gcmdHits: fullURIs.length
-        }),
-        'skos:Concept': concepts
+
+    const [prefLabelMap, shortNameMap, schemeMap, definitionsMap] = await Promise.all([
```
cgokey, line 89: this work only needs to be done if format is json, so move inside the case 'json' block.

serverless/src/getConcept/handler.js
```diff
@@ -68,38 +71,67 @@ export const getConcept = async (event) => {
...
+    const concepts = [
+      { conceptIRI },
+      ...(concept['skos:broader'] ? [{ conceptIRI: concept['skos:broader']['@rdf:resource'] }] : []),
+      ...(concept['skos:narrower'] || []).map((narrow) => ({ conceptIRI: narrow['@rdf:resource'] })),
+      ...(concept['skos:related'] || []).map((related) => ({ conceptIRI: related['@rdf:resource'] }))
+    ]
+
+    const conceptSchemeMap = await createConceptSchemeMap()
+    const prefLabelMap = await createPrefLabelMap(concepts)
```
cgokey, line 92: you don't need to pass in `concepts` to createPrefLabelMap, this function has no arguments.

serverless/src/getConcepts/handler.js
```diff
@@ -88,49 +84,110 @@ export const getConcepts = async (event) => {
...
+      createPrefLabelMap(conceptURIs.map((uri) => ({ conceptIRI: uri }))),
```
cgokey, line 90: createPrefLabelMap takes no arguments.

serverless/src/getConcepts/handler.js
```diff
@@ -88,49 +84,110 @@ export const getConcepts = async (event) => {
...
+      createShortNameMap(),
```
cgokey, line 91: why do you need a shortNameMap and definitions map?  See comment in line 108.   You should just be able to call toLegacyJSON and give it the skos concept, pref label map, and concept scheme map

mandyparson: Calling toLegacyJson requires a skosConcept which requires getSkosConcept which requires an await. When I did that, the program crashed out because I was making too many calls all at once. It's either this solution or process the data in batches which would cause a delay.

serverless/src/getConcept/handler.js
mandyparson: lint doesn't like 'lexical declarations in case block'

serverless/src/getConcepts/handler.js
mandyparson: Same here: "no lexical declaration in case block"

serverless/src/getConcepts/handler.js
cgokey: Sorry, I mispoke... You want something like this:
`
        const skosConcepts = conceptURIs.map((uri) => {
          const ntriples = [...nodes[uri]]

          return toSkosJson(uri, ntriples, bNodeMap)
        })
        const jsonConcepts = skosConcepts.map((concept) => toLegacyJSON(...))
`

serverless/src/getConcepts/handler.js
mandyparson: See comment below about program crashing out with too many calls.

serverless/src/getConcepts/handler.js
```diff
@@ -88,49 +83,107 @@ export const getConcepts = async (event) => {
...
+      ])
+      const jsonResponse = {
+        hits: totalConcepts,
+        page_num: pageNum,
+        page_size: pageSize,
+        termsOfUse: 'https://cdn.earthdata.nasa.gov/conduit/upload/5182/KeywordsCommunityGuide_Baseline_v1_SIGNED_FINAL.pdf',
+        keywordVersion: '20.8',
+        viewer: 'https://gcmd.earthdata.nasa.gov/KeywordViewer/scheme/all',
+        concepts: conceptURIs.map((uri) => {
+          const ntriples = [...nodes[uri]]
```
mandyparson, line 105: change to getLegacyJSON

cgokey (approved):

PR #28, opened 2025-03-13
KMS-535b: Fixing non-default export

mandyparson:
# Overview

### What is the feature?

Bamboo is not building because createChangeNote was not a default export.

### What is the Solution?

Summarize what you changed.

createChangeNote changed from module.exports = createChangeNote to export default createChangeNote

List impacted areas.

# Testing

### Reproduction steps

- **Environment for testing:**
- **Collection to test with:**

1. Step 1
2. Step 2...

### Attachments

# Checklist

- [ ] New and existing unit tests pass locally with my changes
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings

### DIFF
--- a/serverless/src/shared/createChangeNote.js
+++ b/serverless/src/shared/createChangeNote.js
@@ -61,4 +61,4 @@ const createChangeNote = (note) => {
   return changeNote
 }
 
-module.exports = createChangeNote
+export default createChangeNote
--- a/serverless/src/shared/createChangeNote.js
+++ b/serverless/src/shared/createChangeNote.js
@@ -22,7 +22,7 @@
  *     }
  *   }
  */
-const createChangeNote = (note) => {
+export const createChangeNote = (note) => {
   const lines = note.split('\n').map((line) => line.trim())
   const changeNote = {
     changeNoteItems: {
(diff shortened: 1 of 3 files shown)


serverless/src/shared/createChangeNote.js
```diff
@@ -61,4 +61,4 @@ const createChangeNote = (note) => {
   return changeNote
 }
 
-module.exports = createChangeNote
+export default createChangeNote
```
cgokey, line 64: I think you are going to have a problem using this class in the scripts directory now, as it doesn't support ES.

cgokey: looking at this again, I thought this was used in both scripts and main code, but since it is just being used in the main code, you should use a named export, not a default one:
just add the top of your function, add export const createChangeNote, then make sure you import it properly, you'll need the {} around it.

mandyparson: When do we use it in the scripts directory?

htranho (approved):

cgokey (approved):

PR #41, opened 2025-04-21
KMS-560: Code changes for handling export of RDF data from local mysql snapshot

cgokey:
# Overview

### What is the feature?

Pulling all version info from production was too slow and crashed production KMS.   So, what I did was grab a mysql dump, and setup legacy KMS locally to include a snapshot of data and pulled data locally instead.

### What is the Solution?

Wrote scripts such as:
`exportData`, fetches all data from local KMS and writes out the legacy json version of all concepts for each version.   These files are stored in `setup/data/export/json`
`createRdfFiles`, this script turned the legacy json into RDF.   Also produces a schema rdf file as well.
`loadExport`, this script takes the results of createRdfFiles and loads all versions into a RDF file.

Note: `loadExport` will load the RDF data directly to a sparql endpoint.   Therefore, I needed to tunnel into SIT and UAT to load the data.    This is a 1-time thing, so shouldn't be an issue, this is just used to setup the database with the initial data of all version data.

Initial bootstrapping of legacy data into new KMS.

# Testing

In order to test the scripts, you need to have a local version of KMS running.   You can then run all scripts in order:
`npm run export-data` - This connects to a legacy KMS endpoint and produces the legacy json files. (see setup/data/export/json)
`npm run create-rdf-files` - This turns those legacy json files into RDF files. (see setup/data/export/rdf)
`npm run load-export` - This ingests all RDF files into named graphs in the RDF database.

If you want to just verify the end result, as the ticket is "import data into UAT", go here:
https://cmr.uat.earthdata.nasa.gov/kms/concept_versions/version_type/all
And you should see 122 versions.

### Attachments

# Checklist


serverless/src/getConceptVersions/__tests__/handler.test.js
```diff
@@ -67,6 +67,63 @@ describe('getConceptVersions', () => {
...
+
+        expect(response.statusCode).toBe(200)
+        expect(response.headers['Content-Type']).toBe('application/xml; charset=utf-8')
+        expect(response.headers['X-Test']).toBe('test-header')
+
+        const xmlLines = response.body.split('\n').filter((line) => line.trim().startsWith('<version'))
+
+        expect(xmlLines[1]).toBe('  <version type="draft" creation_date="2023-02-01" last_synced="2023-02-02T11:00:00Z">draft</version>')
+        expect(xmlLines[2]).toBe('  <version type="published" creation_date="2023-03-01" last_synced="2023-03-02T12:00:00Z">2.0</version>')
+        expect(xmlLines[3]).toBe('  <version type="published" creation_date="2023-04-01" last_synced="2023-04-02T13:00:00Z">1.5</version>')
```
mandyparson, line 123: two 'published'?

setup/scripts/createRdfFiles.js
```diff
@@ -91,6 +117,7 @@ const createRdfFiles = async () => {
           const json = jsonMap[uuid]
           if (json) {
             const conceptXml = toRDF(json)
+            // eslint-disable-next-line no-await-in-loop
```
mandyparson, line 120: Is there any way to fix this?

setup/scripts/exportData.js
```diff
@@ -0,0 +1,140 @@
+/* eslint-disable no-promise-executor-return */
```
mandyparson, line 1: Again, we should see if we can avoid these

htranho (approved):

setup/scripts/createRdfFiles.js
cgokey: Not easily, and not as readable, I'm not as concerned about these eslint issues since this is being run as a script, and kind of a one-time thing.

setup/scripts/exportData.js
cgokey: Same here, I'm choosing to ignore this linter, as it is much more readable than doing a promise all, and again it is also just a script that is only run once.

serverless/src/getConceptVersions/__tests__/handler.test.js
cgokey: Fixed.

mandyparson (approved):

PR #43, opened 2025-05-09
KMS-576: Improve performance of sparql query for /concepts endpoint

cgokey:
# Overview

### What is the feature?

You can see the issue by going here:
https://cmr.uat.earthdata.nasa.gov/kms/concepts
https://cmr.earthdata.nasa.gov/kms/concepts
Probably one of them will timeout due to the API Gateway having a max timeout of 30 seconds.   If you hit this endpoint multiple times, it will eventually return, probably because the lambda no longer has to do a cold start.

The purpose of this ticket is to improve the performance used to fetch 2000 concepts at a time in order to get the response time well under 30 seconds.

### What is the Solution?

I tracked the slow response time down to 2 major issues, each would significantly degrade performance.
1) The startup of the rdf4j server didn't include a min/max memory heap size, so as a result would leads to suboptimal performance.
2) The sparql query would load ALL triples from the graph then perform the paging subset, so too much data was being loaded into memory.   The new code only loads triples associated with the concept ids in scope (rather than ALL) from the paging parameters.

/kms/concepts endpoint

# Testing

Once this is deployed, compare the performance from:
https://cmr.sit.earthdata.nasa.gov/kms/concepts
https://cmr.uat.earthdata.nasa.gov/kms/concepts
https://cmr.earthdata.nasa.gov/kms/concepts

We should see SIT much faster after these updates.

### Attachments

# Checklist


serverless/src/getConcepts/handler.js
```diff
@@ -105,19 +106,23 @@ export const getConcepts = async (event) => {
...
+        version,
+        pageNum,
+        pageSize
       })
     }
 
     const { bNodeMap, nodes, conceptURIs: fullURIs } = processTriples(triples)
 
-    const totalConcepts = fullURIs.length
+    const conceptURIs = fullURIs
```
cgokey, line 117: I can just use conceptURIs above and don't rename

htranho (approved):

infrastructure/rdfdb/cdk/docker/scripts/setup.sh
```diff
@@ -4,14 +4,17 @@ set -e
 echo 'org.apache.tomcat.util.digester.PROPERTY_SOURCE=org.apache.tomcat.util.digester.EnvironmentPropertySource' >> /usr/local/tomcat/conf/catalina.properties
 echo "RDF4J_DATA_DIR is set to: ${RDF4J_DATA_DIR}"
 
-/usr/local/tomcat/setup_web_auth.sh
+# /usr/local/tomcat/setup_web_auth.sh
```
eudoroolivares2016, line 7: I'm surprised this wasn't erroring out before?

infrastructure/rdfdb/cdk/docker/scripts/setup.sh
```diff
@@ -4,14 +4,17 @@ set -e
...
 
 echo "showing rdf4j data directory"
 ls -l /rdf4j-data
 
-export JAVA_OPTS="-Dorg.eclipse.rdf4j.appdata.basedir=${RDF4J_DATA_DIR}"
-export CATALINA_OPTS="-Dorg.eclipse.rdf4j.appdata.basedir=${RDF4J_DATA_DIR}"
-echo "JAVA_OPTS=\"\$JAVA_OPTS -Dorg.eclipse.rdf4j.appdata.basedir=${RDF4J_DATA_DIR}\"" >> /usr/local/tomcat/bin/setenv.sh
+# Set Java options
+JAVA_OPTS="-Dorg.eclipse.rdf4j.appdata.basedir=${RDF4J_DATA_DIR} -Xmx3g -Xms1g -XX:MaxMetaspaceSize=512m -XX:+UseG1GC -XX:MaxGCPauseMillis=200"
+export JAVA_OPTS
```
eudoroolivares2016, line 14: Can't this just be 
`export JAVA_OPTS="-Dorg.eclipse.rdf4j.appdata.basedir=${RDF4J_DATA_DIR} -Xmx3g -Xms1g -XX:MaxMetaspaceSize=512m -XX:+UseG1GC -XX:MaxGCPauseMillis=200"` ?

serverless/src/getConcepts/__tests__/handler.test.js
```diff
@@ -305,90 +319,134 @@ describe('getConcepts', () => {
...
+      expect(conceptMatches.length).toBe(10)
+
+      for (let i = 90; i < 100; i += 1) {
+        expect(result.body).toContain(`<skos:Concept rdf:about="uri${i}">`)
+        expect(result.body).toContain(`<skos:prefLabel>Concept ${i}</skos:prefLabel>`)
+      }
+
+      expect(getFilteredTriples).toHaveBeenCalledWith({
+        conceptScheme: undefined,
+        pattern: undefined,
```
eudoroolivares2016, line 446: Do we need a test where these patterns are defined I guess thats really just going to be handled on the db side?

serverless/src/shared/getFilteredTriples.js
```diff
@@ -1,70 +1,109 @@
...
+ * @see sparqlRequest - Used to make the SPARQL query requests.
+ * @see getConceptUrisQuery - Used to generate the SPARQL query for fetching concept URIs.
+ * @see getConceptDetailsQuery - Used to generate the SPARQL query for fetching concept details.
  */
 
-export const getFilteredTriples = async ({ conceptScheme, pattern, version }) => {
+export const getFilteredTriples = async ({
+  conceptScheme, pattern, version, pageNum, pageSize
+}) => {
+  const offset = (pageNum - 1) * pageSize
```
eudoroolivares2016, line 65: I'm good with this here but, should we consider having this moved up to the handler that calls this and then passing offset down to here since that value will be constant for any other functions that implement paging on there?

serverless/src/shared/getFilteredTriples.js
```diff
@@ -1,70 +1,109 @@
...
+    const uriResponse = await sparqlRequest({
+      method: 'POST',
+      contentType: 'application/sparql-query',
+      accept: 'application/sparql-results+json',
+      body: uriQuery,
+      version
+    })
+
+    if (!uriResponse.ok) {
+      throw new Error(`HTTP error! status: ${uriResponse.status}`)
```
eudoroolivares2016, line 84:
```suggestion
      throw new Error(`Failed to retrieve concept URIs ! status: ${uriResponse.status}`)
```
just to make it easier to debug later something that references the query that's failing in the log

serverless/src/shared/getFilteredTriples.js
```diff
@@ -1,70 +1,109 @@
...
-        pattern
-      }),
+      body: query,
       version
     })
 
-    if (!response.ok) {
-      throw new Error(`HTTP error! status: ${response.status}`)
+    if (!detailsResponse.ok) {
+      throw new Error(`HTTP error! status: ${detailsResponse.status}`)
```
eudoroolivares2016, line 101:
```suggestion
      throw new Error(`HTTP error! status: ${detailsResponse.status}`)
```
Same with this one. I imagine that this is pretty consistent in the project maybe in the future we can go back and improve the logs a bit to make it more obvious where the error was happening. In the scope of this effort either is fine with me

serverless/src/shared/__tests__/getFilteredTriples.test.js
```diff
@@ -22,106 +24,147 @@ describe('getFilteredTriples', () => {
...
 
-        expect(sparqlRequest).toHaveBeenCalledWith(expect.objectContaining({
-          body: expect.not.stringContaining('skos:prefLabel')
-        }))
+      await getFilteredTriples({
+        conceptScheme: 'sciencekeywords',
+        version: 'published',
+        pageNum: 1,
+        pageSize: 10
       })
```
eudoroolivares2016, line 101: expect(getConceptDetailsQuery).toHaveBeenCalledTimes(1)

eudoroolivares2016: add times check

eudoroolivares2016 (approved):

infrastructure/rdfdb/cdk/docker/scripts/setup.sh
cgokey: Good catch, I think I committed too much

infrastructure/rdfdb/cdk/docker/scripts/setup.sh
cgokey: Yes, changing.

serverless/src/shared/getFilteredTriples.js
cgokey: good suggestion.

serverless/src/shared/getFilteredTriples.js
cgokey: done

serverless/src/shared/__tests__/getFilteredTriples.test.js
cgokey: done

cgokey: done

PR #98, opened 2026-04-08
KMS-659: Pin aws-cdk and aws-cdk-lib to compatible versions for Bamboo deploys

cgokey:
# Overview

### What is the feature?

Fixes Bamboo/local CDK deployment failures caused by an incompatible aws-cdk CLI version relative to the aws-cdk-lib version in the repo.

### What is the Solution?

Pinned aws-cdk and aws-cdk-lib in package.json and updated package-lock.json so the installed CLI is compatible with the cloud assembly schema produced by the library. This resolves the schema version mismatch error during deploy/synth.

Build/deployment tooling
CDK synthesis/deploy path in Bamboo
Local CDK CLI usage from this repo

# Testing

This should now deploy without errors in bamboo.

# Checklist


htranho (approved):

eudoroolivares2016 (approved):
I see this fixed the deployment in bamboo

PR #99, opened 2026-04-09
KMS-660: Publish keyword events to SNS after a successful publish

cgokey:
# Overview

### What is the feature?

Adds the remaining KMS-660 work on top of what is already in main: publish keyword events to SNS after a successful publish, support the downstream local event flow in LocalStack, and harden the local draft/Redis behavior needed to test that flow reliably.

### What is the Solution?

This branch changes only the delta beyond main:

- publisher now publishes generated keyword events to SNS only after the SPARQL publish update succeeds
- post-publish downstream failures are handled as partial-success behavior with explicit logging instead of failing the already-completed publish
- local development now includes a LocalStack bridge that routes:
EventBridge -> publisher
EventBridge -> primeConceptsCache
SNS/SQS -> cmrKeywordEventsListener
- getConcepts draft requests no longer use Redis caching, which prevents stale draft comparisons during local publish testing
- local Redis connection behavior was tightened to fail fast instead of repeatedly retrying in the background
- local startup/docs were updated to support the new bridge workflow 

- Publisher Lambda post-publish behavior
- SNS keyword event publication
- Local LocalStack event routing and testing workflow
- Draft caching behavior in getConcepts
- Local Redis behavior
- Local developer documentation and startup scripts

# Testing

Reproduction steps

Start local dependencies and app:
```
npm run rdf4j:start
npm run localstack:start
npm run redis:start
npm run start-local
```

Create a unique draft keyword:
`./scripts/local/create_unique_keyword.sh`

Publish:
`curl -X POST 'http://127.0.0.1:3013/publish?name=v1.0.0'`

Verify in logs that:
publisher runs
keyword events are published after publish success
primeConceptsCache is invoked
cmrKeywordEventsListener is invoked
Re-run publish without new draft changes and verify the no-op path skips SNS publication.

# Checklist


serverless/src/shared/downloadConcepts.js
```diff
@@ -59,8 +60,13 @@ export const downloadConcepts = async ({ conceptScheme, format = 'csv', version
...
       const errorBody = JSON.parse(result.body)
       const detail = errorBody.error ? ` - ${errorBody.error}` : ''
-      const error = new Error(`Failed to download ${format.toUpperCase()}. Status: ${result.statusCode}${detail}`)
+      const isSchemeNotFound = result.statusCode === 404
+        && (errorBody.error?.includes('Concept scheme not found') ?? false)
+      const error = new Error(
+        `Failed to download ${format.toUpperCase()}. Status: ${result.statusCode}${detail}`
+      )
       error.statusCode = result.statusCode
+      error.isSchemeNotFound = isSchemeNotFound
```
htranho: isSchemeNotFound is not used by the publisher handler.

serverless/src/shared/redisCacheStore.js
```diff
@@ -33,7 +33,9 @@ export const getRedisClient = async () => {
       const client = createClient({
         socket: {
           host: REDIS_HOST,
-          port: REDIS_PORT
+          port: REDIS_PORT,
+          connectTimeout: 5000,
+          reconnectStrategy: () => false
```
eudoroolivares2016: This seems like it'd affect prod as well. I guess this is fine if it fails to connect to redis then it'll just fail and not try again

README.md
```diff
@@ -51,35 +56,11 @@ Local development intentionally splits responsibilities between SAM and LocalSta
 - LocalStack emulates AWS-managed services that SAM does not model end-to-end for this repo, especially SNS and SQS.
 - RDF4J and Redis remain separate local services because they are not AWS services.
 
-We do not run the entire application stack inside LocalStack because the existing SAM flow is simpler for day-to-day Lambda/API development, while LocalStack is most useful here for the managed messaging pieces. For keyword event processing, `npm run start-local` also starts [`scripts/local/run_localstack_cmr_keyword_events_bridge.js`], which polls the LocalStack SQS queue and forwards messages into the local CMR consumer handler. This bridge exists because `sam local start-api` does not emulate SQS event source mappings the way AWS does in deployed environments.
+We do not run the entire application stack inside LocalStack because the existing SAM flow is simpler for day-to-day Lambda/API development, while LocalStack is most useful here for the managed messaging pieces. For keyword event processing, `npm run start-local` also starts `scripts/localstack/run_bridge.sh`, which runs `scripts/localstack/bridge.js`.
 
-### Testing keyword event publishing in SIT
```
eudoroolivares2016, line 56: Can we also remove the `serverless/src/keywordEventsTestPublish/handler.js` now it can wait otherwise too if we wanna use it for running tests in cloud

serverless/src/shared/downloadConcepts.js
```diff
@@ -59,8 +60,13 @@ export const downloadConcepts = async ({ conceptScheme, format = 'csv', version
     if (result.statusCode !== 200) {
       const errorBody = JSON.parse(result.body)
       const detail = errorBody.error ? ` - ${errorBody.error}` : ''
-      const error = new Error(`Failed to download ${format.toUpperCase()}. Status: ${result.statusCode}${detail}`)
+      const isSchemeNotFound = result.statusCode === 404
+        && (errorBody.error?.includes('Concept scheme not found') ?? false)
```
eudoroolivares2016: I don't see issue with this but, is there a case where that would happen? From keyword manager it really can't happen unless someone tries to use the API to publish and includes a a wrong schema?

serverless/src/publisher/handler.js
```diff
@@ -367,9 +454,45 @@ export const publisher = async (event) => {
...
+        )
+
+        publishSummary.failedEvents.forEach(({ keywordEvent, error, attempts }) => {
+          logger.error(
+            '[publisher] Keyword event publish exhausted retries',
+            {
+              uuid: keywordEvent.UUID,
+              scheme: keywordEvent.Scheme,
+              eventType: keywordEvent.EventType,
+              attempts,
```
htranho, line 493: Maybe also add the paths (old + new) in the log, convenient in case we need to do something.

scripts/localstack/bridge.js
```diff
@@ -0,0 +1,574 @@
...
+const main = async () => {
+  configureHostHandlerEnvironment()
+  const bridgeRegistry = getBridgeRegistry()
+
+  const shuttingDownRef = { value: false }
+  process.on('SIGINT', () => {
+    shuttingDownRef.value = true
+  })
+
+  process.on('SIGTERM', () => {
```
eudoroolivares2016: I think this is just a duplicate

serverless/src/publisher/handler.js
```diff
@@ -241,19 +247,38 @@ export const getKeywordChanges = async () => {
...
+
+    return sequentialResults
+  }, Promise.resolve([]))
 
   // Collect all successful comparisons into a Map
   const keywordChangesMap = new Map(
     results
-      .filter((result) => result.status === 'fulfilled' && result.value)
-      .map((result) => [result.value.notation, result.value.comparison])
+      .filter(Boolean)
```
eudoroolivares2016: This will pass over rejected results which might be fine but, maybe a comment as to why

serverless/src/publisher/handler.js
```diff
@@ -241,19 +247,38 @@ export const getKeywordChanges = async () => {
...
+  if (failedSchemes.length > 0) {
+    const failedSchemeSummary = failedSchemes
+      .map(({ notation, error }) => `${notation}: ${error}`)
+      .join('; ')
+
+    throw new Error(
+      `Keyword changes detection failed for ${failedSchemes.length} `
+      + `scheme(s): ${failedSchemeSummary}`
+    )
+  }
```
eudoroolivares2016: This is where I think we add the feature toggle

serverless/src/publisher/handler.js
```diff
@@ -295,23 +320,86 @@ const emitCachingEvent = async ({ versionName, publishDate, keywordEvents }) =>
...
+ * @async
+ * @param {Array<Object>} keywordEvents - Keyword event payloads to publish.
+ * @returns {Promise<{
+ *   attemptedCount: number,
+ *   publishedCount: number,
+ *   failedEvents: Array<{ keywordEvent: Object, error: string, attempts: number }>
+ * }>}
+ * SNS publish summary for the completed batch.
+ */
+const publishKeywordEvents = async (keywordEvents) => keywordEvents.reduce(
```
eudoroolivares2016, line 349: we should break this lambda up soon its doing alot of different methods

serverless/src/publisher/handler.js
```diff
@@ -367,9 +454,45 @@ export const publisher = async (event) => {
...
+
+    if (keywordEvents.length > 0) {
+      const publishSummary = await publishKeywordEvents(keywordEvents)
+      keywordEventsPublished = publishSummary.publishedCount
+
+      if (publishSummary.failedEvents.length > 0) {
+        warnings.push(
+          `Publish completed, but ${publishSummary.failedEvents.length} `
+          + 'keyword event publishes failed after retries'
+        )
```
eudoroolivares2016: nitpick; feels like all these aren't really warnings right I mean its failing to do something

serverless/src/shared/downloadConcepts.js
cgokey: Double checked, you are right, I'm going to remove it.

serverless/src/publisher/handler.js
cgokey: Adding...

serverless/src/shared/downloadConcepts.js
cgokey: Mostly agree. In the normal publisher flow this shouldn’t happen, because we fetch the draft/published scheme lists first and only download schemes from that computed set. The only realistic cases are a direct bad API call or the data changing between the scheme-list read and the per-scheme download. Since publisher wasn’t actually using this flag anymore, I removed it.

scripts/localstack/bridge.js
cgokey: Good call. I collapsed those into a shared shutdown handler and registered it for both SIGINT and SIGTERM.

serverless/src/publisher/handler.js
cgokey: Good call. I simplified this instead: we now throw on any failed schemes before building the map, so there’s no longer a need to filter out null/rejected placeholders here.

serverless/src/publisher/handler.js
cgokey: Added a feature toggle here: BLOCK_PUBLISH_ON_KEYWORD_DIFF_FAILURE. It defaults to non-blocking for now, so diff failures log and continue with successful scheme comparisons until we’re ready to make that path live.

serverless/src/publisher/handler.js
cgokey: Agreed, a follow-up refactor should split publisher into smaller units for diff analysis, publish execution, SNS publish, and cache-prime emission.

serverless/src/publisher/handler.js
cgokey: Agreed. I renamed the result field from `warnings` to `postPublishFailures` so it better reflects that these are real downstream failures after publish, not just warnings.

README.md
cgokey: Removed

serverless/src/shared/redisCacheStore.js
cgokey: Good catch. I scoped the fail-fast Redis socket settings behind REDIS_FAIL_FAST, and only enable that in local env. So local bridge runs still fail fast, but prod/non-local keeps the default Redis reconnect behavior.

htranho (approved):

scripts/localstack/bridge.js
eudoroolivares2016: Oh honestly didn't notice they were for different shutdown commands

eudoroolivares2016 (approved):

PR #101, opened 2026-04-13
KMS-654: Ensure KMS logs from the lambdas are available in Splunk

htranho:
# Overview

### What is the feature?

Ensure KMS logs from the lambdas are available in Splunk

### What is the Solution?

1. Added new helper class, LogForwardingSetup.ts, that:
- Creates CloudWatch Log Groups for each Lambda function with retention policy
- Sets up Subscription Filters to forward logs to NGAP SecLog destination
- Uses cross-account destination ARN (no IAM role needed)
- Validates required parameters
2. KmsLambdaFunctions.ts
- Add stage name to function names
3. KmsStack.ts
- Add logDestinationAccount and secLogAccount to interface.
- Add LogForwardingSetup
4. CmrEventProcessingStack.js
- Add logDestinationAccount and secLogAccount to interface.
- Add LogForwardingSetup
5. main.ts
- Read new env vars, added validation, pass to stacks.
6. deploy_bamboo.sh
- Added 2 new environment variables to Docker run

Logging now forwarded to Splunk. Example for sit env:
log_group="/aws/lambda/kms-sit*" 

# Testing

Invoke kms endpoints, use log_group above in Splunk to search for logs.

### Attachments

N/A

# Checklist

- [ ] I have made corresponding changes to the documentation


cdk/app/lib/helper/LogForwardingSetup.ts
```diff
@@ -0,0 +1,119 @@
...
+    const { lambdas } = props
+
+    // Set up log forwarding for each Lambda function
+    Object.entries(lambdas).forEach(([key, lambdaFunction]) => {
+      const sanitizedKey = key.replace(/[^a-zA-Z0-9]/g, '-')
+
+      // Create explicit log group with retention
+      // Log group naming: /aws/lambda/{prefix}-{stage}-{functionName}
+      // This enables easy Splunk searches: source="/aws/lambda/kms-sit-*"
+      const logGroup = new logs.LogGroup(scope, `${sanitizedKey}-LogGroup`, {
```
cgokey, line 96: Did you deploy this already?   I know AWS automatically creates log groups, does this override the existing log groups already created?  I'm wondering if you'll get conflicts trying to explicity create a log group with a retention policy that was automatically already created.

cgokey: maybe something like this: ```const logRetention = new logs.LogRetention(scope, `${sanitizedKey}-LogRetention`, {
      logGroupName,
      retention: this.logRetentionDays
    })```. then you are not recreating the log group.

cdk/app/lib/KmsStack.ts
```diff
@@ -160,6 +170,21 @@ export class KmsStack extends cdk.Stack {
...
 
+    // Set up CloudWatch Logs forwarding to Splunk via NGAP SecLog account
+    // Skip log forwarding for localstack deployments
+    if (!useLocalstack) {
+      // Create log forwarding setup - no reference needed as it registers itself
+      // eslint-disable-next-line no-new
+      new LogForwardingSetup(this, 'LogForwarding', {
+        prefix,
+        stage,
+        account: logDestinationAccount || this.account,
```
cgokey: I think this may be over-parameterized. If the destination name is always keyed by the current/source account, shouldn’t this just be `this.account`? Carrying a separate `logDestinationAccount` here feels redundant unless we have a real case where that suffix differs from the current account.`

cdk/app/lib/helper/KmsLambdaFunctions.ts
```diff
@@ -617,7 +617,7 @@ export class LambdaFunctions {
     let lambdaFunction = this.lambdas[lambdaKey]
     if (!lambdaFunction) {
       const nodejsFunctionProps: NodejsFunctionProps = {
-        functionName: `${this.props.prefix}-${functionName}`,
+        functionName: `${this.props.prefix}-${this.props.stage}-${functionName}`,
```
cgokey, line 620: One thing I’d like to double-check here: changing the function name to `${prefix}-${stage}-${functionName}` looks like it will replace all existing KMS Lambdas on deploy, not just add log forwarding. Was that intentional for this ticket? If not, it may be worth avoiding the rename so we don’t widen the rollout impact of a logging change.`

cgokey: I can see the value in this even though it forces a one-time Lambda replacement. It gives us consistent stage-aware naming and makes the log-group/log-forwarding story cleaner.

cdk/bin/main.ts
```diff
@@ -68,6 +70,14 @@ async function main() {
...
   }
 
+  // Validate required logging parameters for non-localstack deployments
+  if (!useLocalstack && !secLogAccount) {
+    throw new Error(
+      'SEC_LOG_ACCOUNT environment variable is required for log forwarding. '
+      + 'Please set bamboo_SEC_LOG_ACCOUNT in your Bamboo deployment configuration.'
+    )
+  }
+
```
eudoroolivares2016: The deployment would just failout right if we miss one of the env vars?

cdk/app/lib/KmsStack.ts
```diff
@@ -160,6 +170,21 @@ export class KmsStack extends cdk.Stack {
...
+        region: this.region,
+        secLogAccount: secLogAccount!,
```
cgokey: Since `SEC_LOG_ACCOUNT` is already validated in `main.ts` for non-localstack deploys, could we make it required in the stack/construct props as well instead of carrying it as optional and asserting with `!`? I think that would make the contract a little clearer and keep the typing aligned with the actual runtime requirement.

cdk/app/lib/helper/LogForwardingSetup.ts
```diff
@@ -0,0 +1,119 @@
...
+        'secLogAccount is required for log forwarding. '
+        + 'Please set the SEC_LOG_ACCOUNT environment variable in your Bamboo deployment.'
+      )
+    }
+
+    this.secLogAccount = props.secLogAccount
+
+    // Set destination ARN using the account IDs from Bamboo deployment parameters
+    // The destination is in the SecLog account but accepts logs from our account
+    this.destinationName = `arn:aws:logs:${props.region}:${this.secLogAccount}:destination:/gsfc-ngap-managed/application_logs_destination/${props.account}`
```
eudoroolivares2016: I think this can just be one env var. @htranho look in bamboo for graphql for instance and you'll see the `LOG_DESTINATION_ARN` then we wouldn't need two diff env vars for this

cdk/app/lib/KmsStack.ts
htranho: Now using logDestinationArn.

cdk/app/lib/KmsStack.ts
htranho: Now using logDestinationArn

cdk/bin/main.ts
htranho: Now using logDestinationArn. Yes, the deployment will fail if that param not set.

cdk/app/lib/helper/LogForwardingSetup.ts
htranho: Now using logDestinationArn.

cdk/bin/main.ts
```diff
@@ -187,7 +197,8 @@ async function main() {
     prefix,
     stage,
     stackName: `${prefix}-CmrEventProcessingStack`,
-    topicArn: kmsStack.keywordEventsTopic.topicArn
+    topicArn: kmsStack.keywordEventsTopic.topicArn,
+    logDestinationArn: logDestinationArn || ''
```
htranho: this too

cdk/app/lib/helper/LogForwardingSetup.ts
```diff
@@ -0,0 +1,108 @@
...
+  private readonly logRetentionDays: logs.RetentionDays
+
+  /**
+   * Constructs a new LogForwardingSetup instance
+   * @param {Construct} scope - The scope in which to define this construct
+   * @param {string} id - The construct ID
+   * @param {LogForwardingSetupProps} props - Configuration properties
+   */
+  constructor(scope: Construct, id: string, props: LogForwardingSetupProps) {
+    this.logRetentionDays = props.logRetentionDays || logs.RetentionDays.ONE_WEEK
```
eudoroolivares2016: Do we need retention period here? I don't think we use one in other other apps at all in the cdk lib `https://git.earthdata.nasa.gov/projects/EDSC/repos/edsc-cdk-utils/browse/src/application/nodeJsFunction.ts#118`

cdk/app/lib/helper/LogForwardingSetup.ts
```diff
@@ -0,0 +1,108 @@
...
+        retention: this.logRetentionDays,
+        removalPolicy: cdk.RemovalPolicy.DESTROY
+      })
+
+      // Create subscription filter to forward logs to NGAP SecLog destination
+      // The destination ARN is pre-configured in the SecLog account and has
+      // the necessary permissions to accept logs from our account
+      const subscriptionFilter = new logs.CfnSubscriptionFilter(
+        scope,
+        `${sanitizedKey}-SplunkSubscription`,
```
eudoroolivares2016: nitpick;
```suggestion
        `${sanitizedKey}-subscriptionFilter`,
```
I don't wanna include `Splunk` in the resource in case later we migrate to ELK its fine if comments in the code but, the actual resource name we should just keep unattached to a service

eudoroolivares2016:
```suggestion
           new logs.CfnSubscriptionFilter(scope, `${capitalizeFirstLetter(functionName)}SubscriptionFilter`, {
      destinationArn: logDestinationArn,
      filterPattern: '',
      logGroupName: logGroup.logGroupName
    })
```
Ideally we could just match the pattern name in the cdk lib

eudoroolivares2016 (approved):

cdk/app/lib/helper/LogForwardingSetup.ts
htranho: Now set to one month

cgokey (approved):

PR #102, opened 2026-04-16
KMS-661: Add Monitoring and Metrics for event publishing (KMS)

cgokey:
# Overview

### What is the feature?

Adds keyword sync observability to publisher by emitting CloudWatch metrics for keyword change detection and SNS keyword event publishing. This also adds a CloudWatch alarm for publish failures and optional email notifications when that alarm fires.

### What is the Solution?

This branch adds keyword sync metrics under the `KMS/KeywordSync` namespace and emits them from the publisher flow during three points in the publish lifecycle:

- after keyword diff analysis:
  - `KeywordChangesDetected`
- after keyword event creation:
  - `KeywordEventsGenerated`
- after SNS publish attempts complete:
  - `KeywordEventsPublished`
  - `KeywordEventPublishFailures`

The implementation includes:

- a new shared metric helper for emitting CloudWatch metrics from publisher
- publisher changes to count and emit metrics without making metric emission itself block publish
- IAM updates to allow `cloudwatch:PutMetricData`
- LocalStack CloudWatch support so the metrics path can be exercised locally
- a CloudWatch alarm on `KeywordEventPublishFailures`
- optional SNS email subscriptions for that alarm via `KEYWORD_SYNC_ALARM_EMAILS`
- Bamboo/CDK wiring for the new monitoring configuration

- Publisher runtime and tests
- Shared serverless metric utilities
- KMS IAM/CDK infrastructure
- LocalStack startup for local verification
- Bamboo deploy-time environment plumbing
- 
# Testing

1. Local verification
 Start LocalStack and local KMS:
 ```bash
 npm run localstack:start
 npm run start-local
./scripts/local/create_unique_keyword.sh
curl -X POST 'http://127.0.0.1:3013/publish?name=v1.0.0'
aws --endpoint-url=http://localhost:4566 cloudwatch list-metrics --namespace KMS/KeywordSync
./scripts/local/show_keyword_sync_metrics.js
```
2. AWS verification
Create or update a draft keyword in the target environment
Publish through the normal AWS flow
In CloudWatch, open KMS/KeywordSync -> Metrics with no dimensions
Verify these metric names appear and receive datapoints:
KeywordChangesDetected
KeywordEventsGenerated
KeywordEventsPublished
KeywordEventPublishFailures
Use Statistic = Sum and a time window that includes the publish run

3. Alarm verification
If KEYWORD_SYNC_ALARM_EMAILS is configured, confirm SNS email subscriptions are confirmed
You will get a email confirming you want this subscription as well, so look for that.
```
aws cloudwatch set-alarm-state \
  --alarm-name kms-sit-keyword-event-publish-failures \
  --state-value ALARM \
  --state-reason "Manual KMS-661 email test"
```

# Checklist

### DIFF
--- a/bin/deploy-bamboo.sh
+++ b/bin/deploy-bamboo.sh
@@ -64,6 +64,7 @@ dockerRun() {
         --env "EDL_PASSWORD=$bamboo_EDL_PASSWORD" \
         --env "CMR_BASE_URL=$bamboo_CMR_BASE_URL" \
         --env "BLOCK_PUBLISH_ON_KEYWORD_DIFF_FAILURE=${bamboo_BLOCK_PUBLISH_ON_KEYWORD_DIFF_FAILURE:-false}" \
+        --env "KEYWORD_SYNC_ALARM_EMAILS=${bamboo_KEYWORD_SYNC_ALARM_EMAILS:-}" \
         --env "CORS_ORIGIN=$bamboo_CORS_ORIGIN" \
         --env "RDF4J_INSTANCE_TYPE=$bamboo_RDF4J_INSTANCE_TYPE" \
         --env "RDF4J_CONTAINER_MEMORY_LIMIT=$bamboo_RDF4J_CONTAINER_MEMORY_LIMIT" \
--- a/cdk/app/lib/helper/KeywordSyncMonitoringSetup.ts
+++ b/cdk/app/lib/helper/KeywordSyncMonitoringSetup.ts
@@ -1,0 +1,64 @@
+import * as cdk from 'aws-cdk-lib'
+import * as cloudwatch from 'aws-cdk-lib/aws-cloudwatch'
+import * as cloudwatchActions from 'aws-cdk-lib/aws-cloudwatch-actions'
+import * as sns from 'aws-cdk-lib/aws-sns'
+import * as subscriptions from 'aws-cdk-lib/aws-sns-subscriptions'
+import { Construct } from 'constructs'
+
+const KEYWORD_SYNC_METRIC_NAMESPACE = 'KMS/KeywordSync'
+const KEYWORD_EVENT_PUBLISH_FAILURES_METRIC = 'KeywordEventPublishFailures'
+
+export interface KeywordSyncMonitoringSetupProps {
+  prefix: string
+  stage: string
+  notificationEmails?: string[]
+}
+
+/**
+ * Sets up CloudWatch monitoring resources for keyword sync publishing.
+ */
+export class KeywordSyncMonitoringSetup {
+  public readonly publishFailuresAlarm: cloudwatch.Alarm
+
+  /**
+   * Creates the keyword sync monitoring resources.
+   *
+   * @param {Construct} scope - Construct scope for the monitoring resources.
+   * @param {KeywordSyncMonitoringSetupProps} props - Monitoring configuration.
+   */
+  constructor(scope: Construct, props: KeywordSyncMonitoringSetupProps) {
+    const { notificationEmails = [], prefix, stage } = props
+
+    const keywordEventPublishFailuresMetric = new cloudwatch.Metric({
+      namespace: KEYWORD_SYNC_METRIC_NAMESPACE,
+      metricName: KEYWORD_EVENT_PUBLISH_FAILURES_METRIC,
+      statistic: 'Sum',
+      period: cdk.Duration.minutes(5)
+    })
+
+    this.publishFailuresAlarm = new cloudwatch.Alarm(scope, 'KeywordSyncPublishFailuresAlarm', {
+      alarmName: `${prefix}-${stage}-keyword-event-publish-failures`,
+      alarmDescription: 'Alerts when publisher keyword event publishes fail after retries.',
+      metric: keywordEventPublishFailuresMetric,
+      threshold: 1,
+      evaluationPeriods: 1,
+      datapointsToAlarm: 1,
+      comparisonOperator: cloudwatch.ComparisonOperator.GREATER_THAN_OR_EQUAL_TO_THRESHOLD,
+      treatMissingData: cloudwatch.TreatMissingData.NOT_BREACHING
+    })
+
+    if (notificationEmails.length === 0) {
+      return
+    }
+
+    const alarmTopic = new sns.Topic(scope, 'KeywordSyncAlarmTopic', {
+      topicName: `${prefix}-${stage}-keyword-sync-alerts`
+    })
+
+    notificationEmails.forEach((email) => {
+      alarmTopic.addSubscription(new subscriptions.EmailSubscription(email))
+    })
+
+    this.publishFailuresAlarm.addAlarmAction(new cloudwatchActions.SnsAction(alarmTopic))
+  }
+}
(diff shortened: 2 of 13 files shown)

### CODE REVIEW

File: serverless/src/shared/emitPublisherMetrics.js [hunk 1]
```diff
@@ -0,0 +1,137 @@
...
+ * tests use this form while deployed AWS continues to use the SDK client.
+ *
+ * @param {Object} params - Query serialization details.
+ * @param {Array<{metricName: string, value: number}>} params.metrics - Metrics to emit.
+ * @returns {string} URL-encoded request body for PutMetricData.
+ */
+const buildCloudWatchQueryRequestBody = ({ metrics }) => {
+  const params = new URLSearchParams({
+    Action: 'PutMetricData',
+    Version: '2010-08-01',
```
cgokey, line 58: why this?

cdk/app/lib/helper/KeywordSyncMonitoringSetup.ts
```diff
@@ -0,0 +1,64 @@
+import * as cdk from 'aws-cdk-lib'
+import * as cloudwatch from 'aws-cdk-lib/aws-cloudwatch'
+import * as cloudwatchActions from 'aws-cdk-lib/aws-cloudwatch-actions'
+import * as sns from 'aws-cdk-lib/aws-sns'
+import * as subscriptions from 'aws-cdk-lib/aws-sns-subscriptions'
+import { Construct } from 'constructs'
+
+const KEYWORD_SYNC_METRIC_NAMESPACE = 'KMS/KeywordSync'
+const KEYWORD_EVENT_PUBLISH_FAILURES_METRIC = 'KeywordEventPublishFailures'
```
eudoroolivares2016, line 9: Should we add an alarm to this so we get informed if this part of the pipeline fails?

cdk/app/lib/helper/KeywordSyncMonitoringSetup.ts
```diff
@@ -0,0 +1,64 @@
...
+   * @param {KeywordSyncMonitoringSetupProps} props - Monitoring configuration.
+   */
+  constructor(scope: Construct, props: KeywordSyncMonitoringSetupProps) {
+    const { notificationEmails = [], prefix, stage } = props
+
+    const keywordEventPublishFailuresMetric = new cloudwatch.Metric({
+      namespace: KEYWORD_SYNC_METRIC_NAMESPACE,
+      metricName: KEYWORD_EVENT_PUBLISH_FAILURES_METRIC,
+      statistic: 'Sum',
+      period: cdk.Duration.minutes(5)
```
eudoroolivares2016: This might be noisy at 5 minutes
`cdk.Duration.days(1)` 
Shouldn't it match the alarm time anyways?

bin/deploy-bamboo.sh
```diff
@@ -64,6 +64,7 @@ dockerRun() {
         --env "EDL_PASSWORD=$bamboo_EDL_PASSWORD" \
         --env "CMR_BASE_URL=$bamboo_CMR_BASE_URL" \
         --env "BLOCK_PUBLISH_ON_KEYWORD_DIFF_FAILURE=${bamboo_BLOCK_PUBLISH_ON_KEYWORD_DIFF_FAILURE:-false}" \
+        --env "KEYWORD_SYNC_ALARM_EMAILS=${bamboo_KEYWORD_SYNC_ALARM_EMAILS:-}" \
```
eudoroolivares2016, line 67: Feels like if we are gonna do this we should use the support email which will go to Zendesk platform. I don't think KMS has a direct one but, we could have one made or otherwise use MMTs at least for prod

serverless/src/shared/emitPublisherMetrics.js
eudoroolivares2016: This is just the API version for cloudwatch right?  saw the same value in one of the scripts `scripts/local/show_keyword_sync_metrics.js`

serverless/src/publisher/handler.js
```diff
@@ -14,10 +15,54 @@ const KEYWORD_EVENT_PUBLISH_RETRIES = 3
...
+ * @returns {number} Total keyword changes detected for the publish run.
+ */
+const countKeywordChanges = (keywordChangesMap) => Array.from(keywordChangesMap.values()).reduce(
+  (total, changes) => total
+    + changes.addedKeywords.size
+    + changes.removedKeywords.size
+    + changes.changedKeywords.size,
+  0
+)
+
```
eudoroolivares2016, line 44:
```suggestion
const countKeywordChanges = (keywordChangesMap) => Array.from(keywordChangesMap.values()).reduce(
  (total, { addedKeywords, removedKeywords, changedKeywords }) => total
  + addedKeywords.size
  + removedKeywords.size
  + changedKeywords.size,
  0
)
```
I tried a for of which I thought was easier to read too but, the linter complained

serverless/src/publisher/handler.js
```diff
@@ -441,16 +485,38 @@ export const publisher = async (event) => {
...
 
+    await emitPublisherMetricsSafely(
+      [
+        {
+          metricName: PUBLISHER_METRIC_NAMES.KEYWORD_EVENTS_GENERATED,
+          value: keywordEventsGenerated
+        }
+      ],
+      'keyword event generation'
+    )
```
eudoroolivares2016: should we just batch these together to have a single call only?

bin/deploy-bamboo.sh
cgokey: Yes, that makes sense to me.

cdk/app/lib/helper/KeywordSyncMonitoringSetup.ts
cgokey: Yep, this PR adds a CloudWatch alarm for KeywordEventPublishFailures; it notifies through SNS when KEYWORD_SYNC_ALARM_EMAILS is configured.

serverless/src/publisher/handler.js
cgokey: Good suggestion.  Made this change

serverless/src/shared/emitPublisherMetrics.js
cgokey: Yes, for this path. This is the CloudWatch Query API version, not a KMS/versioning value. The deployed AWS path uses the SDK and does not need it, but the LocalStack fallback sends the classic Query API request shape directly, where Version=2010-08-01 is required.

cdk/app/lib/helper/KeywordSyncMonitoringSetup.ts
cgokey: Good call. I changed the failure alarm to aggregate over a one-day period instead of five minutes, and pulled the period into a named constant so the intended alarm window is clearer.

serverless/src/publisher/handler.js
cgokey: Good call. I batched these into a single metric emission after the SNS publish summary is known, so all four keyword sync metrics are sent together in one CloudWatch call.

htranho (approved):

cdk/app/lib/helper/KeywordSyncMonitoringSetup.ts
eudoroolivares2016: Oh I forgot to delete this comment I realized that early on 👍

eudoroolivares2016 (approved):

PR #103, opened 2026-04-20
KMS-663: Export Historical KMS Keyword Versions

htranho:
# Overview

### What is the feature?

Export Historical KMS Keyword Versions

### What is the Solution?

Create scripts to:
1. Download past version's rdf files from S3.
2. Process locally (upload to local rdf4j, use kms to download csv) to get csv files for each scheme.
3. Upload csv files to S3.

When publishing a new keyword version, rdf file and csv files of the new version are uploaded to S3.

Past versions processing.
Publishing new version.

# Testing

1. Use config to specify a version(s) to download from S3 (Edit scripts-config.sh). Run 'npm run download-rdf'. Downloaded files should be in 'archive-processor/downloaded-rdf/'
2. Test change some keyword label.
3. Start up local env (rdf4j, kms)
5. Run 'npm run process-rdf' to upload the rdf from step 1 and then download the produced csv from local kms. Downloaded files should be in 'archive-processor/local-kms-csv/[version]/'
6. Verify changes made in step 2.
7. Run 'npm run upload-csv' to upload to S3.

Use MMT to publish a new version, then check AWS S3 to verify rdf and csv files uploaded in the new version directory in S3.

### Attachments

N/A

# Checklist


archive-processor/scripts/scripts-config.sh
```diff
@@ -0,0 +1,73 @@
...
+# RDF Downloader Settings (for download-rdf-from-S3.js)
+# =============================================================================
+
+# Delay in milliseconds between downloads to avoid rate limiting.
+export DOWNLOAD_DELAY_MS="100"
+
+# Optional: Comma-separated list of specific versions to download.
+# If this string is empty, the script will download all RDF files.
+# Example: export TO_BE_DOWNLOADED_VERSIONS="10.0,11.0,KMS-654-Testing"
+export TO_BE_DOWNLOADED_VERSIONS="7.0,8.0"
```
htranho: Add instruction for case of all versions

eudoroolivares2016:
1. Make the CSV headers which we don't have data for use `NA` string
2. Do we have an automated solution to update the consolidated CSVs on publish?

archive-processor/scripts/scripts-config.sh
htranho: Done:
# If this string is empty, the script will download all RDF files.

archive-processor/scripts/process-rdf.js
```diff
@@ -0,0 +1,507 @@
...
+
+  if (versionTagRegex.test(xmlData)) {
+    xmlData = xmlData.replace(versionTagRegex, newVersionTag)
+    console.log(`Replaced version tag. New tag: ${newVersionTag}`)
+  } else {
+    // This is not a fatal error, as some schemes files may not have this tag.
+    console.warn(`Warning: Could not find <gcmd:versionName> tag in ${filePath}. Uploading file as-is.`)
+  }
+
+  // Also set created date to 'N/A'. In the scheme csv output, this is 'Revision'
```
htranho, line 284: This scheme file doesn't control 'Timestamp' in the csv, but the 'Revision'

htranho: Checked: Timestamp is current time when csv created in kms.

htranho: Now 'Revision' set to 'N/A'

htranho:
> 1. Make the CSV headers which we don't have data for use `NA` string
> 2. Do we have an automated solution to update the consolidated CSVs on publish?

1. Yes
2. Yes, rdf and csv files are uploaded to S3 at publishing.

serverless/src/publisher/handler.js
```diff
@@ -507,6 +510,36 @@ export const publisher = async (event) => {
       logger.info('[publisher] No keyword events generated, skipping SNS publish')
     }
```
cgokey, line 560: I’m okay with keeping archival export inline as long as this handler consistently finishes well below the 15-minute Lambda timeout.

Since publish/SNS side effects happen before the archival exports, a timeout here would not be catchable and could leave cache-prime un-emitted or cause an EventBridge retry after some downstream work already happened.

Can we add a strong code comment near the archival export section calling out that this work must stay comfortably under the Lambda timeout? If we ever get close to that limit, we should move archival export to separate async work or reconsider emitting cache-prime before archival export.

serverless/src/shared/exportPublishSchemeCsvToS3.js
```diff
@@ -0,0 +1,79 @@
...
+
+        if (s3ExportDelayMs > 0 && index < schemes.length - 1) {
+          await delay(s3ExportDelayMs)
+        }
+      } catch (error) {
+        logger.error(`Failed to process scheme ${notation}: ${error.message}`)
+      }
+    }), Promise.resolve())
+
+    logger.info('Finished exporting all published scheme CSVs to S3.')
```
cgokey, line 82: I like that this keeps exporting the remaining schemes if one scheme fails. Could we also collect the failed scheme notations and report them back to the caller at the end?

Right now a per-scheme failure is logged here, but the helper still resolves successfully and the publisher logs `Successfully exported Published Scheme CSVs to S3.` That could make a partial historical export look like a full success unless someone notices the earlier error log.

```suggestion
const failedSchemes = []

// inside per-scheme catch:
failedSchemes.push({ notation, error })

// after loop:
if (failedSchemes.length > 0) {
  throw new Error(`Failed to export CSV for schemes: ${failedSchemes.map(({ notation }) => notation).join(', ')}`)
}
```

bin/localstack/start.sh
```diff
@@ -6,7 +6,7 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 # shellcheck source=bin/env/local_env.sh
 source "${SCRIPT_DIR}/../env/local_env.sh"
 
-REQUIRED_SERVICES="sns,sqs,events"
+REQUIRED_SERVICES="sns,sqs,events,s3"
```
cgokey: When this branch rebases on current `main`, can we make sure this ends up including both new LocalStack services?

I think the final value should be:

```bash
REQUIRED_SERVICES="sns,sqs,events,s3,cloudwatch"

File: archive-processor/scripts/scripts-config.sh [hunk 6]
```diff
@@ -0,0 +1,73 @@
...
+# =============================================================================
+# RDF Downloader Settings (for download-rdf-from-S3.js)
+# =============================================================================
+
+# Delay in milliseconds between downloads to avoid rate limiting.
+export DOWNLOAD_DELAY_MS="100"
+
+# Optional: Comma-separated list of specific versions to download.
+# If this string is empty, the script will download all RDF files.
+# Example: export TO_BE_DOWNLOADED_VERSIONS="10.0,11.0,KMS-654-Testing"
```
cgokey, line 63: The README says these version lists are optional filters and that leaving them empty processes all available versions, but the checked-in config defaults to only `7.0,8.0`.

For this historical export workflow, I think the safer default is to process all versions:

```bash
export TO_BE_DOWNLOADED_VERSIONS=""
export TO_BE_PROCESSED_VERSIONS=""
export TO_BE_UPLOADED_VERSIONS=""

File: archive-processor/scripts/download-rdf-from-S3.js [hunk 7]
```diff
@@ -0,0 +1,299 @@
...
+ * AWS Profile to use (optional)
+ * @type {string|undefined}
+ */
+const awsProfile = process.env.AWS_PROFILE
+
+/**
+ * Output directory for downloaded files
+ * @type {string}
+ */
+const outputDir = join(scriptDir, '..', 'downloaded-rdf')
```
cgokey, line 35: Can we add the generated archive output directories to `.gitignore`?

These scripts write local RDF/CSV artifacts under the repo:

```text
archive-processor/downloaded-rdf/
archive-processor/local-kms-csv/

File: archive-processor/scripts/download-rdf-from-S3.js [hunk 8]
```diff
@@ -0,0 +1,299 @@
...
+ * These should be the S3 key prefixes (e.g., "10.0", "KMS-654-Testing").
+ * @type {string}
+ */
+const toBeDownloadedVersions = process.env.TO_BE_DOWNLOADED_VERSIONS || ''
+
+/**
+ * S3 Client configured for the specified region
+ * @type {S3Client}
+ */
+const s3Client = new S3Client({
```
cgokey: Do these archive scripts support LocalStack S3?

I see this branch adds `s3` to the LocalStack services, but these standalone archive scripts create `new S3Client({ region })` directly. The app-side S3 client uses the shared `awsClients` helper, which wires `AWS_ENDPOINT_URL`, test credentials, and `forcePathStyle` for LocalStack.

If local S3 support is intended here too, can we add endpoint/path-style config to these scripts, probably using `AWS_ENDPOINT_URL=http://localhost:4566` for host-run scripts?

bin/localstack/start.sh
htranho: Done

archive-processor/scripts/scripts-config.sh
htranho: Done

archive-processor/scripts/download-rdf-from-S3.js
htranho: Done

serverless/src/publisher/handler.js
htranho: Done

serverless/src/shared/exportPublishSchemeCsvToS3.js
htranho: Done, added code to consider scheme error.

archive-processor/scripts/download-rdf-from-S3.js
htranho: Now added support for local stack, which can be set in script config file.

archive-processor/scripts/process-rdf.js
```diff
@@ -0,0 +1,507 @@
...
+
+/**
+ * Clears a graph context in the repository for a given version.
+ * @param {string} version - Graph version (for example `published` or `draft`).
+ * @returns {Promise<void>}
+ * @throws {Error} If context delete fails.
+ */
+const clearContext = async (version) => {
+  console.log(`Clearing context: ${version}`)
+  const graphUri = `https://gcmd.earthdata.nasa.gov/kms/version/${version}`
```
eudoroolivares2016, line 152:
```suggestion
  const graphUri = `https://cmr.earthdata.nasa.gov/kms/version/${version}`
```

eudoroolivares2016: Isn't this the old endpoint on KMS 1.0

serverless/src/shared/awsClients.js
```diff
@@ -0,0 +1,50 @@
...
+      region: process.env.AWS_REGION || 'us-east-1',
+      credentials: {
+        accessKeyId: 'test',
+        secretAccessKey: 'test'
+      }
+    }
+    : {}
+
+  if (endpoint) {
+    config.forcePathStyle = true
```
eudoroolivares2016, line 19:
```suggestion
    // // Use path-style URLs (e.g. host/bucket/key) instead of subdomain-style (e.g. bucket.host/key)
// Required for local S3-compatible services like MinIO and LocalStack
    config.forcePathStyle = true
```

This is for localstack to work right?

archive-processor/scripts/process-rdf.js
```diff
@@ -0,0 +1,507 @@
...
+ * @param {string} filePath - Local path to RDF/XML file.
+ * @param {string} version - Context version (for example `published` or `draft`).
+ * @returns {Promise<void>}
+ * @throws {Error} If file read or POST load fails.
+ */
+const loadRdfFile = async (filePath, version) => {
+  const xmlData = await readFile(filePath, 'utf8')
+  console.log(`Read ${xmlData.length} bytes from file: ${filePath}`)
+
+  const graphUri = `https://gcmd.earthdata.nasa.gov/kms/version/${version}`
```
eudoroolivares2016, line 239: same with this one?

serverless/src/shared/__tests__/exportRdfToS3.test.js
```diff
@@ -0,0 +1,139 @@
...
+      text: vi.fn().mockResolvedValue('<rdf>Test RDF Data</rdf>')
+    })
+
+    S3Client.prototype.send = vi.fn().mockResolvedValue({})
+    ensureBucketAndLifecycleRule.mockResolvedValue()
+  })
+
+  afterEach(() => {
+    vi.useRealTimers()
+  })
```
eudoroolivares2016:
```suggestion

```
not needed took it off locally and passed I don't think we are really  it

serverless/src/shared/__tests__/exportPublishSchemeCsvToS3.test.js
```diff
@@ -0,0 +1,227 @@
...
+    // Mocking resolved values for each download call
+    downloadConcepts.mockImplementation(async ({ conceptScheme }) => {
+      if (conceptScheme === 'SCHEME1') return 'csv,data,for,scheme1'
+      if (conceptScheme === 'SCHEME2') return 'csv,data,for,scheme2'
+
+      return ''
+    })
+  })
+
+  afterEach(() => {
```
eudoroolivares2016:
```suggestion

```
I don't think this is needed right because we just need to useFake timers for the setTimeout once per test?

bin/localstack/start.sh
```diff
@@ -8,7 +8,7 @@ source "${SCRIPT_DIR}/../env/local_env.sh"
 
 # CloudWatch is enabled here so keyword sync metrics can be verified locally
 # alongside the existing SNS/SQS/EventBridge workflow.
-REQUIRED_SERVICES="sns,sqs,events,cloudwatch"
+REQUIRED_SERVICES="sns,sqs,events,s3,cloudwatch"
 
 if ! docker network inspect "${KMS_DOCKER_NETWORK}" >/dev/null 2>&1; then
   docker network create "${KMS_DOCKER_NETWORK}" >/dev/null
```
cgokey, line 14: on line 29, we shoudl put this instead now:
```
if [[ ",${configured_services}," != *",events,"* ]] \
  || [[ ",${configured_services}," != *",s3,"* ]] \
  || [[ ",${configured_services}," != *",cloudwatch,"* ]]; then
```

cgokey (approved):
I had one more minor comment but then I think it looks good.

eudoroolivares2016 (approved):

bin/localstack/start.sh
htranho: yes, done

archive-processor/scripts/process-rdf.js
htranho: All uri, resource in rdf are using gcmd.earthdata... We will need to change it in future work if we want?

serverless/src/shared/awsClients.js
htranho: Yes it is

archive-processor/scripts/process-rdf.js
htranho: See comment above.

serverless/src/shared/__tests__/exportRdfToS3.test.js
htranho: Correct, now removed

serverless/src/shared/__tests__/exportPublishSchemeCsvToS3.test.js
htranho: Now removed

PR #105, opened 2026-04-30
KMS-664: Build historical keywords cache and create new endpoints to access

htranho:
# Overview

### What is the feature?

Build historical keywords cache and create new endpoints to access

### What is the Solution?

1. New API Endpoints:

- GET /concept/historical/full_path/{fullPath+}, mandatory parameter 'scheme'
- GET /concept/historical/short_name/{shortName}, mandatory parameter 'scheme'

2. Create cache builder which is run by the publishing process to go over all versions, process all CSVs to fill the Redis cache which historical keywords using full path or short name as key.

Publishing new keyword version.

# Testing

### Reproduction steps

1. Verify Cache Build Process

- Trigger a publish event in the application. Example:
curl -X POST 'http://127.0.0.1:3013/publish?name=50.0'
- After the publisher handler completes, check its console logs or CloudWatch logs.
- Expected Result: You should see a log message stating, "Successfully built Historical Concept cache from S3 bucket...".

2. Test the historical/full_path Endpoint

Schemes have keyword cached with fullPath as key: 
[
  'sciencekeywords',
  'locations',
  'chronounits',
  'rucontenttype',
  'isotopiccategory',
  'temporalresolutionrange',
  'horizontalresolutionrange'
] 

- Make a GET request to the new endpoint using a known historical full path.
- Example: GET /concept/historical/full_path/EARTH+SCIENCE+%3E+OCEANS?scheme=sciencekeywords
- Expected Result: The request should return a 200 OK with a JSON body containing the correct uuid and fullPath.

4. Test the historical/short_name Endpoint

Schemes have keyword cached with shortName as key: 
[
  'providers',
  'platforms',
  'instruments',
  'projects',
  'idnnode',
  'DataFormat'
] 

- Make a GET request to the new endpoint using a known historical short name.
- Example: GET /concept/historical/short_name/AC-690A?scheme=instruments
- Expected Result: The request should return a 200 OK with a JSON body containing the correct uuid, fullPath and longName (if any).

'providers' has different CSV structure, also test to verify
- Example: GET /concept/historical/short_name/BGSU%2FGEOL?scheme=providers
- Expected Result: The request should return a 200 OK with a JSON body containing the correct uuid, fullPath and longName (if any).

### Attachments

N/A

# Checklist

### DIFF
--- a/serverless/src/getHistoricalConceptByFullPath/handler.js
+++ b/serverless/src/getHistoricalConceptByFullPath/handler.js
@@ -1,4 +1,6 @@
-import { DEFAULT_FULL_PATH_SCHEMES } from '@/shared/constants/fullPathForUuidSchemes'
+import {
+  DEFAULT_HISTORICAL_CONCEPT_FULL_PATH_SCHEMES
+} from '@/shared/constants/fullPathForHistoricalConceptSchemes'
 import { getApplicationConfig } from '@/shared/getConfig'
 import { logAnalyticsData } from '@/shared/logAnalyticsData'
 import { logger } from '@/shared/logger'
@@ -6,11 +8,11 @@ import { createConceptResponseCacheKeyByFullPath } from '@/shared/redisCacheKeys
 import { getCachedJsonResponse } from '@/shared/redisCacheStore'
 
 /**
- * Retrieves a cached UUID and fullPath for a given fullPath. This handler ONLY checks the cache
+ * Retrieves a historical concept (UUID and fullPath) for a given fullPath from the cache. This handler ONLY checks the cache
  * and will return a 404 if the value is not already cached.
  *
  * @async
- * @function getCachedConceptByFullPath
+ * @function getHistoricalConceptByFullPath
  * @param {Object} event - The Lambda event object.
  * @param {Object} event.pathParameters - The path parameters from the API Gateway event.
  * @param {string} event.pathParameters.fullPath - The full path of the concept.
@@ -33,18 +35,18 @@ import { getCachedJsonResponse } from '@/shared/redisCacheStore'
  *   body: '{\"uuid\":\"91697b7d-8f2b-4954-850e-61d5f61c867d\",\"fullPath\":\"EARTH SCIENCE > OCEANS\"}'
  * };
  */
-export const getCachedConceptByFullPath = async (event, context) => {
+export const getHistoricalConceptByFullPath = async (event, context) => {
   const {
     defaultResponseHeaders,
-    schemesForUuidByFullPath: schemesFromConfig
+    schemesForHistoricalConceptByFullPath: schemesFromConfig
   } = getApplicationConfig()
 
   const sourceForSchemes = (schemesFromConfig && schemesFromConfig.length > 0)
     ? schemesFromConfig
-    : DEFAULT_FULL_PATH_SCHEMES
-  const schemesForUuidByFullPath = sourceForSchemes.map((s) => s.toLowerCase())
+    : DEFAULT_HISTORICAL_CONCEPT_FULL_PATH_SCHEMES
+  const schemesForHistoricalConceptByFullPath = sourceForSchemes.map((s) => s.toLowerCase())
 
-  logger.debug(`Using schemes for fullPath cache: ${JSON.stringify(schemesForUuidByFullPath)}`)
+  logger.debug(`Using schemes for fullPath cache: ${JSON.stringify(schemesForHistoricalConceptByFullPath)}`)
 
   logAnalyticsData({
     event,
@@ -78,7 +80,7 @@ export const getCachedConceptByFullPath = async (event, context) => {
     }
   }
 
-  if (!schemesForUuidByFullPath.includes(scheme)) {
+  if (!schemesForHistoricalConceptByFullPath.includes(scheme)) {
     return {
       statusCode: 400,
       headers: {
@@ -100,7 +102,7 @@ export const getCachedConceptByFullPath = async (event, context) => {
 
     const cachedResponse = await getCachedJsonResponse({
       cacheKey,
-      entityLabel: 'Concept by fullPath'
+      entityLabel: 'Historical Concept by fullPath'
     })
 
     if (cachedResponse) {
@@ -118,7 +120,7 @@ export const getCachedConceptByFullPath = async (event, context) => {
       body: JSON.stringify({ error: 'Cached Concept not found for the given fullPath' })
     }
   } catch (error) {
-    logger.error(`Error retrieving cached UUID, error=${error.toString()}`)
+    logger.error(`Error retrieving historical concept, error=${error.toString()}`)
 
     return {
       headers: defaultResponseHeaders,
@@ -130,4 +132,4 @@ export const getCachedConceptByFullPath = async (event, context) => {
   }
 }
 
-export default getCachedConceptByFullPath
+export default getHistoricalConceptByFullPath
--- a/serverless/src/getHistoricalConceptByFullPath/handler.js
+++ b/serverless/src/getHistoricalConceptByFullPath/handler.js
@@ -1,5 +1,5 @@
 import {
-  DEFAULT_HISTORICAL_CONCEPT_FULL_PATH_SCHEMES
+  HISTORICAL_CONCEPT_FULL_PATH_SCHEMES
 } from '@/shared/constants/fullPathForHistoricalConceptSchemes'
 import { getApplicationConfig } from '@/shared/getConfig'
 import { logAnalyticsData } from '@/shared/logAnalyticsData'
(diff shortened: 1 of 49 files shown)


serverless/src/shared/baseConceptCacheBuilder.js
```diff
@@ -0,0 +1,192 @@
...
+
+    // Use Redis pipeline for batch writes
+    const BATCH_SIZE = 1000
+    let totalWritten = 0
+
+    // Process batches sequentially to avoid overwhelming Redis with concurrent pipelines
+    /* eslint-disable no-await-in-loop */
+    for (let i = 0; i < cacheEntries.length; i += BATCH_SIZE) {
+      const batch = cacheEntries.slice(i, i + BATCH_SIZE)
+      const pipeline = redisClient.multi()
```
cgokey: One thing that might be worth trying for performance is replacing multi().set(...).exec()with mSet per batch (or plain pipelining). We’re already batching 1000 writes at a time, so the next likely win is reducing the transaction/per-command overhead fromMULTI/EXEC, especially since this cache build doesn’t seem to require transactional semantics for each batch.
```suggestion
      const keyValuePairs = batch.flatMap(({ key, value }) => [key, value])

      try {
        await redisClient.mSet(keyValuePairs)
        totalWritten += batch.length
        logger.debug(
          `[cache-builder] Wrote batch progress=${totalWritten}/${cacheEntries.length} `
          + `scheme=${options.scheme}`
        )
      } catch (error) {
        logger.error(
          `[cache-builder] Pipeline batch failed scheme=${options.scheme} `
          + `batch=${Math.floor(i / BATCH_SIZE) + 1} error=${error.message}`
        )
      }
```

serverless/src/shared/buildHistoricalConceptCache.js
```diff
@@ -0,0 +1,198 @@
...
+  const PROCESS_BATCH_SIZE = 5
+  logger.info(`Processing files in parallel batches of ${PROCESS_BATCH_SIZE}.`)
+
+  let successCount = 0
+  let failCount = 0
+
+  const results = await processBatch(
+    allCsvFiles,
+    async (key) => {
+      const scheme = path.basename(key, '.csv').toLowerCase()
```
cgokey, line 205: Minor perf thought: since scheme is derived from the S3 key, could we check whether the scheme is supported before downloading the object? Right now we still doGetObject for unsupported CSVs, even though we end up warning and skipping them.

serverless/src/getHistoricalConceptByFullPath/handler.js
```diff
@@ -0,0 +1,129 @@
...
+ * // with UUID '91697b7d-8f2b-4954-850e-61d5f61c867d', the handler returns:
+ * const response = {
+ *   statusCode: 200,
+ *   headers: {
+ *     'Content-Type': 'application/json'
+ *   },
+ *   body: '{\"uuid\":\"91697b7d-8f2b-4954-850e-61d5f61c867d\",\"fullPath\":\"EARTH SCIENCE > OCEANS\"}'
+ * };
+ */
+export const getHistoricalConceptByFullPath = async (event, context) => {
```
cgokey, line 38: Should we add these two new historical lookup endpoints to getCapabilitiesas well? They’re routed inKmsLambdaFunctions.ts, but right now they aren’t discoverable from the capabilities response.

serverless/src/publisher/handler.js
```diff
@@ -583,6 +589,7 @@ export const publisher = async (event) => {
       logger.error(`[publisher] ${failureMessage}`)
     }
 
+    // Export published schemes CSVs to S3
     try {
       await exportPublishSchemeCsvToS3()
```
cgokey, line 614: buildHistoricalConceptCache() probably shouldn’t run if exportPublishSchemeCsvToS3() failed. In that case we’d be rebuilding Redis from stale or partial S3 data, which could make the new historical lookup endpoints serve outdated mappings right after publish.

serverless/src/shared/buildHistoricalConceptCache.js
```diff
@@ -0,0 +1,198 @@
...
+
+  const fullPathCacheBuilder = new ConceptForFullPathCacheBuilder()
+  const shortNameCacheBuilder = new ConceptForShortNameCacheBuilder()
+
+  /**
+   * Lists the top-level directories in the S3 bucket, which correspond to keyword versions.
+   * @returns {Promise<string[]>} A promise that resolves to an array of version directory prefixes.
+   */
+  const listVersionDirectories = async () => {
+    const command = new ListObjectsV2Command({
```
cgokey, line 91: Could we at least add a warning comment here that this only reads the first ListObjectsV2 page? It’s fine for now, but we’d need to revisit it if we ever hit more than 1000 top-level version folders or more than 1000 objects under a single version prefix.

serverless/src/shared/buildHistoricalConceptCache.js
```diff
@@ -0,0 +1,198 @@
...
+      Bucket: bucketName,
+      Key: key
+    })
+    logger.debug(`Downloading content from [s3://${bucketName}/${key}].`)
+    const response = await s3Client.send(command)
+
+    return streamToString(response.Body)
+  }
+
+  logger.info(`Starting cache build from S3 bucket [${bucketName}].`)
```
cgokey, line 153: Since this is taking ~264s, it may be worth adding a little phase timing here too (list directories, list files, download CSV, parse, Redis write). That would make it much easier to tell whether the real bottleneck is S3, parsing, or Redis before tuning batch sizes further.

serverless/src/getHistoricalConceptByShortName/handler.js
```diff
@@ -0,0 +1,133 @@
...
+ * // and longName 'Aerocommander aircraft', the handler returns:
+ * const response = {
+ *   statusCode: 200,
+ *   headers: {
+ *     'Content-Type': 'application/json'
+ *   },
+ *   body: '{\"uuid\":\"6fa682b9-c6b5-46ca-971f-b7ecd4bf304d\",\"fullPath\":\"Air-based Platforms > Propeller > AC-690A\",\"longName\":\"Aerocommander aircraft\"}'
+ * };
+ */
+export const getHistoricalConceptByShortName = async (event, context) => {
```
cgokey, line 42: Should we add the two new historical lookup endpoints to getCapabilities as well? They’re routed in KmsLambdaFunctions.ts, but right now they aren’t discoverable from the capabilities response.

serverless/src/shared/conceptForShortNameCacheBuilder.js
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  /**
+   * Parses CSV content and returns a map of short names to UUIDs.
+   * @param {string} csvContent - CSV content as a string.
+   * @param {object} options
+   * @param {string} options.scheme - The scheme ('instruments', 'providers', etc.).
+   * @returns {Map<string, {uuid: string, fullPath: string, longName: string}>} Map with short name as key and an object with uuid, fullPath, and longName as value.
+   */
+  parseCsvContent(csvContent, { scheme }) {
```
eudoroolivares2016, line 52: maybe an improvement for later but, it would be nice to see if we can remove some of the hardcoded elements and just have a module that searches the CSV by the column name

serverless/src/shared/baseConceptCacheBuilder.js
```diff
@@ -0,0 +1,188 @@
...
+
+      try {
+        await redisClient.mSet(keyValuePairs)
+        totalWritten += batch.length
+        logger.debug(
+          `[cache-builder] Wrote batch progress=${totalWritten}/${cacheEntries.length} `
+          + `scheme=${options.scheme}`
+        )
+      } catch (error) {
+        logger.error(
```
cgokey, line 110: Could we let Redis batch write failures propagate here instead of only logging them? Right now an mSet error still makes processToCache() resolve, so buildHistoricalConceptCache() treats the file as successfully processed and publish can report success while the historical cache is incomplete. Even rethrowing here, or aggregating failed batches and throwing once at the end, would make the failure visible to the caller.

serverless/src/shared/buildHistoricalConceptCache.js
```diff
@@ -0,0 +1,232 @@
...
+  if (versionDirs.length === 0) {
+    logger.warn('No version directories found in the bucket. Nothing to process.')
+
+    return
+  }
+
+  // Phase 2: List CSV files in all directories
+  phaseStartTime = Date.now()
+  const LIST_BATCH_SIZE = 5
+  const allCsvFiles = (await processBatch(
```
cgokey: Could we fail the rebuild if any per-directory S3 listing rejects here? Right now we filter allSettled() down to fulfilled results and continue, which can silently omit an entire version directory from the historical cache. Since this cache is supposed to represent all archived versions, I think a visible rebuild failure is safer than a partial cache that looks successful.

serverless/src/shared/buildHistoricalConceptCache.js
```diff
@@ -0,0 +1,232 @@
...
+    },
+    PROCESS_BATCH_SIZE
+  )
+  phaseTimes.processFiles = ((Date.now() - phaseStartTime) / 1000).toFixed(2)
+
+  results.forEach((result, index) => {
+    if (result.status === 'fulfilled') {
+      successCount += 1
+    } else {
+      failCount += 1
```
cgokey: One other thing I’d consider tightening here: if any CSV file fails during download/parse/cache processing, we currently just log it and keep going, but the overall build still resolves successfully. That means publisher can report a successful historical cache build even when part of the archive was skipped. Could we throw when failCount > 0 (ideally with the failed keys/reasons) so partial cache builds are surfaced as real failures?

cgokey (approved):

eudoroolivares2016 (approved):

PR #108, opened 2026-05-08
KMS-675:  Consumer - Implement Metadata Correction Service abstract class 

cgokey:
# Overview

### What is the feature?

KMS-675 adds end-to-end metadata correction for keyword-driven collection remediation.

After a keyword publish changes or deletes a published keyword, KMS can now:

- discover which CMR collections still reference the affected keyword UUID
- validate supported collection keywords against the current published KMS keyword set
- resolve historical keyword values back to the correct concept UUID
- map that UUID to the current published replacement keyword path, or determine that the keyword should be deleted
- dispatch the correction to a native-format delegate
- persist one audit record per resolved correction

This branch also adds the local smoke-test toolkit and cache wiring needed to exercise that flow reliably.

### What is the Solution?

This branch adds the metadata-correction workflow and the Redis cache strategy that supports it.

Key changes:

- Adds the metadata-correction event-processing path in CDK, including:
  - CMR keyword event listener wiring
  - metadata correction request queue/consumer wiring
  - Redis / RDF4J / CMR environment configuration for the correction Lambdas
  - VPC/security-group configuration so the correction services can reach Redis and RDF4J
- Updates `cmrKeywordEventsListener` to:
  - consume keyword events
  - look up affected collection concept ids in CMR
  - publish one correction request per impacted collection
- Implements `metadataCorrectionService` to:
  - fetch collection UMM details from CMR
  - validate supported keyword fields instead of relying on CMR keyword validation freshness
  - extract keyword validation failures in a CMR-like response shape
  - resolve historical keyword values through Redis-backed full-path and short-name lookups
  - resolve replacement keyword paths from a published Redis UUID index
  - apply corrections through format-specific delegates
  - persist an RDF audit record for each resolved correction
- Adds the metadata-correction audit read API and capabilities entry so correction history can be queried by collection, keyword UUID, scheme, action, and status.
- Extends publish-time cache handling so:
  - published keyword lookup caches are rebuilt before keyword events are emitted
  - historical caches are rebuilt before keyword events are emitted
  - publisher fetches published concepts CSV directly from the source of truth, bypassing the normal published response cache, so stale /concepts cache entries cannot be used to rebuild the published correction caches
- Extends published Redis priming to write a UUID-indexed lookup in addition to the existing full-path and short-name lookups, which removes RDF4J from replacement-path resolution.
- Improves historical cache building by tracking which immutable S3 historical versions have already been fully written to Redis, so later rebuilds can skip them.
- Adds the local KMS-675 smoke-test toolkit:
  - mock CMR server
  - keyword-event publisher
  - historical + published cache seeder
  - audit-log viewer
  - checked-in end-to-end fixture
- Removes the old dependency on mock/CMR ingest validation for the smoke flow. Keyword validity now comes from the published Redis cache.
- Documents the Redis cache key families, value shapes, and cache-clearing behavior in the README.

Current delegate status:

- UMM-C corrections are implemented and exercised in the local smoke flow.
- DIF10 / ECHO10 / ISO19115 / ISO-SMAP currently remain handoff stubs that preserve the correction contract while real mutation work is follow-on.

- CDK event-processing infrastructure
- Publisher publish/caching flow
- CMR keyword event listener
- Metadata correction service
- Metadata correction audit API
- Historical keyword lookup APIs
- Shared Redis cache builders and lookup helpers
- Shared CMR fetch/validation/extraction/resolution helpers
- Local smoke-test scripts and fixtures
- README operational/cache documentation

# Testing

1. Start local dependencies:

```bash
npm run rdf4j:start
npm run rdf4j:setup
npm run redis:start
npm run localstack:start
```

2. Start the local API:

```bash
npm run start-local
```

3. In a separate terminal, start the mock CMR:

```bash
bash scripts/local/start_mock_cmr_server.sh
```

4. In another terminal, start the LocalStack bridge with metadata-correction local overrides:

```bash
REDIS_ENABLED=true \
REDIS_HOST=localhost \
REDIS_PORT=6380 \
RDF4J_SERVICE_URL=http://localhost:8081 \
RDF4J_USER_NAME=rdf4j \
RDF4J_PASSWORD=rdf4j \
USE_LOCALSTACK=true \
CMR_BASE_URL=http://127.0.0.1:3020 \
MOCK_CMR_BASE_URL=http://127.0.0.1:3020 \
MOCK_CMR_WRITEBACK_ENABLED=true \
LOCALSTACK_HOST_ENDPOINT=http://localhost:4566 \
bash scripts/localstack/run_bridge.sh
```

5. Seed the historical and published keyword caches:

```bash
npx vite-node --config vite.config.js \
  scripts/local/seed_metadata_correction_keyword_caches.js
```

6. Publish the fixture keyword events:

```bash
node scripts/local/publish_keyword_event_local.mjs
```

7. Inspect the audit rows:

```bash
node scripts/local/show_metadata_correction_audit_log.mjs
```

Expected local smoke story:

- first `UPDATED sciencekeywords` event:
  - resolves and applies the invalid science keyword
  - resolves and applies the invalid platform short name
- second `UPDATED platforms` event:
  - sees only the remaining delete-only instrument issue
  - does not remove the instrument yet
- final `DELETED instruments` event:
  - resolves the delete-only instrument match
  - removes the instrument from metadata

Expected audit rows from the checked-in smoke fixture:

- `sciencekeywords UPDATED applied`
- `platforms UPDATED applied`
- `instruments DELETED applied`

# Checklist


bin/deploy-bamboo.sh
```diff
@@ -3,6 +3,8 @@
 # Bail on unset variables, errors and trace execution
 set -eux
 
+: "${bamboo_CMR_BASE_URL:?bamboo_CMR_BASE_URL is required}"
```
cgokey: remove

htranho (approved):

serverless/src/cmrKeywordEventsListener/__tests__/handler.test.js
```diff
@@ -79,13 +131,188 @@ describe('when the CMR keyword events processor is invoked', () => {
...
+        )
+
+        expect(getCmrCollectionConceptIds).not.toHaveBeenCalled()
+        expect(publishMetadataCorrectionRequest).not.toHaveBeenCalled()
+        expect(result).toEqual({
+          batchItemFailures: []
+        })
+      })
+
+      test('should log n/a publish metadata when the publish helper omits optional fields', async () => {
```
eudoroolivares2016: what is this use-case?

serverless/src/cmrKeywordEventsListener/handler.js
```diff
@@ -46,30 +87,105 @@ const buildMetadataCorrectionRequest = (keywordEvent) => ({
...
+    let keywordPath
+
     try {
-      const snsEnvelope = JSON.parse(record.body || '{}')
+      // Unwrap the SNS envelope first, then parse the original KMS keyword event payload.
+      const {
+        body,
+        messageId: recordMessageId
+      } = record
+      messageId = recordMessageId
```
eudoroolivares2016: can we just set it once

eudoroolivares2016:
```suggestion
      messageId =  messageId = ''
```

serverless/src/cmrKeywordEventsListener/handler.js
```diff
@@ -46,30 +87,105 @@ const buildMetadataCorrectionRequest = (keywordEvent) => ({
...
+          logger.info(
+            '[consumer] No affected collection concept ids found for keyword event '
+            + `scheme=${scheme} `
+            + `uuid=${uuid} `
+            + `keywordPath=${keywordPath || 'n/a'}`
+          )
+        }
+
+        // Publish a collection-scoped correction request for each affected concept id.
+        await Promise.all(collectionConceptIds.map(async (collectionConceptId) => {
```
eudoroolivares2016: I wanna double check we aren't doing too many callbacks in here

eudoroolivares2016 (approved):

serverless/src/cmrKeywordEventsListener/__tests__/handler.test.js
cgokey: This isn’t meant to represent a normal real-world success case. Normally we expect the success log to look like:

`[consumer] Published metadata correction request collectionConceptId=C1000000000-PROV messageId=metadata-correction-message-123 topicArn=arn:aws:sns:us-east-1:000000000000:kms-dev-metadata-correction-requests.fifo`

This test is only exercising the defensive `|| 'n/a'` defaults so that if part of the returned publish metadata is missing, the log still renders cleanly as:

`[consumer] Published metadata correction request collectionConceptId=C1000000000-PROV messageId=n/a topicArn=n/a`

If we only cover the messageId case, branch coverage drops because the topicArn || 'n/a' fallback is no longer exercised.

serverless/src/cmrKeywordEventsListener/handler.js
cgokey: Good catch. We need messageId in the outer scope for error logging, but we can initialize it once from record?.messageId and just destructure body inside the try. I’ll clean that up.

serverless/src/cmrKeywordEventsListener/handler.js
cgokey: I pulled the inner publish loop into a small helper so the handler no longer inlines a nested Promise.all(...map(async ...)) block. That improves readability only; it does not change the concurrency behavior. We still publish collection-scoped correction requests in parallel for each affected concept id, but the downstream queue is FIFO and publishMetadataCorrectionRequest() uses collectionConceptId as the MessageGroupId, so correction requests for the same collection are still serialized in order. If the concern is readability, this refactor addresses that; if the concern is overall concurrency volume, that would be a separate issue.

PR #109, opened 2026-05-19
KMS-672: Update keywords in DIF10 native record

htranho:
# Overview

### What is the feature?

Update keywords in DIF10 native record.

### What is the Solution?

Implement corrections for keywords in DIF10 records.

Keywords in DIF10 correction

# Testing

### Reproduction steps

run test applyDif10MetadataCorrections.test.js, this test based on a DIF10 in mocks directory: Modify the corrections in the last test to check the corrected DIF10.

### Attachments

N/A

# Checklist


cgokey:
This looks like a meaningful step on the DIF10 updater itself, but I don’t think the current branch satisfies the AC around actually updating records in the consumer flow and reindexing them.
Right now metadataCorrectionService/handler.js is still a placeholder that only logs incoming requests, so the new DIF10 logic in applyDif10MetadataCorrections.js is not yet wired into the real correction path. Because of that, the acceptance criterion about updated records being reingested/reindexed feels out of scope for what this branch actually delivers.
My suggestion would be:
remove the AC related to consumer-side application/reingest from this ticket
track the real end-to-end consumer wiring in a follow-up ticket
put that eventual runtime behavior behind a feature toggle before enabling it in shared environments
That would let this ticket stay focused on the DIF10 field-mutation implementation itself, while keeping the riskier “live consumer updates native records and reindexes them” behavior isolated and controllable when it is introduced.

serverless/src/shared/applyDif10MetadataCorrections.js
```diff
@@ -0,0 +1,136 @@
...
+  indentBy: '    ',
+  suppressEmptyNode: true
+})
+
+/**
+ * Mapping of UMM correction schemes to their respective delegate functions.
+ * Each delegate is responsible for performing targeted mutations on the parsed XML object.
+ * @type {Object.<string, Function>}
+ */
+const SCHEME_DELEGATES = {
```
cgokey: I think we still have a coverage gap against the DIF10 controlled keyword fields here: productlevelid does not appear to be implemented.
The DIF10 updater map in applyDif10MetadataCorrections.js includes handlers for science keywords, platforms, instruments, projects, providers, locations, chrono units, RU content type, ISO topic category, IDN node, and resolution ranges, but I don’t see a delegate for productlevelid. The DIF10 fixture also still contains `<Product_Level_Id>`, so this looks like a real native field we would need to mutate for that scheme.
As written, a productlevelid correction would likely be ignored rather than updated, which seems inconsistent with the ticket scope of updating DIF10 controlled keyword fields.
Can we either:
add productlevelid handling and tests in this branch, or
explicitly narrow the supported DIF10 scheme list in the ticket/PR so this omission is intentional?

serverless/src/shared/__tests__/applyDif10MetadataCorrections.test.js
```diff
@@ -0,0 +1,199 @@
...
+      metadataPayload: mockDif10,
+      corrections: []
+    })
+
+    expect(result.correctedMetadata).toContain('<?xml version="1.0" encoding="UTF-8"?>')
+    expect(result.correctedMetadata).toContain('<DIF>')
+  })
+})
+
+describe('Correct a DIF10', () => {
```
cgokey: I think the higher-level DIF10 fixture test needs stronger assertions, because right now it may be masking a missed correction.
In applyDif10MetadataCorrections.test.js, the returns a corrected DIF10 test builds a realistic mixed correction set, but it only logs the result instead of asserting the expected outcome. 
So while the per-scheme unit tests are strong, this combined “realistic DIF10 fixture” test is not currently protecting us against one of the included corrections silently failing.
Could we tighten this test to assert at least:
the expected correctionCount
the exact set of applied schemes
one or more concrete XML assertions for each correction family included in the fixture
That would make this test much more valuable as an integration-style check instead of just a debug print.

serverless/src/shared/Dif10FieldsCorrection/applyProviderCorrection.js
```diff
@@ -0,0 +1,79 @@
...
+  if (!targetOrg) return false
+
+  // Ensure Organization_Name exists for replacement
+  if (!targetOrg.Organization_Name && action === 'replace') {
+    targetOrg.Organization_Name = {}
+  }
+
+  const target = targetOrg.Organization_Name
+
+  if (action === 'delete') {
```
cgokey: I want to sanity-check the delete semantics for DIF10 blocks that contain more than just the controlled keyword fields.

For example, in applyProviderCorrection.js, a delete removes the entire <Organization> block, not just Organization_Name. That also drops non-keyword metadata attached to that block, such as hours of service, instructions, and personnel/contact information.

Not sure if there are other cases like this as well, but we should double check.

serverless/src/shared/Dif10FieldsCorrection/applyScienceKeywordCorrection.js
```diff
@@ -0,0 +1,75 @@
...
+  } else if (index === 0) {
+    target = keywords
+  }
+
+  if (!target) return false
+
+  if (action === 'delete') {
+    const parent = parsedMetadata.DIF
+    if (Array.isArray(parent.Science_Keywords)) {
+      parent.Science_Keywords.splice(index, 1)
```
cgokey: I think this is a place where matching by the actual keyword value would be safer than trusting the original ummPath index.
As corrections are applied sequentially to one mutable parsed XML object, and helpers like applyScienceKeywordCorrection.js can splice() array entries on delete. After that, later corrections are still using the original index from ummPath, even though the array may have shifted.

serverless/src/shared/applyDif10MetadataCorrections.js
```diff
@@ -0,0 +1,136 @@
...
+  }, Promise.resolve({
+    finalMetadata: initialParsedMetadata,
+    applied: []
+  }))
+
+  return {
+    ...params,
+    correctionCount: applied.length,
+    // Reconstruct the XML payload with the standard UTF-8 declaration
+    correctedMetadata: `<?xml version="1.0" encoding="UTF-8"?>\n${xmlBuilder.build(finalMetadata)}`,
```
cgokey: I’m a little concerned that this implementation rewrites the entire DIF10 XML document on every correction, rather than making a narrowly scoped edit. In applyDif10MetadataCorrections.js, we parse the full document and rebuild the full document, and the parser/builder settings include trimValues: true and suppressEmptyNode: true. That means even a targeted keyword update can normalize whitespace, change mixed-text formatting, or drop empty nodes elsewhere in the record. I’m not saying this is definitely wrong, but if we’re going to take the full parse/rebuild approach, I think the tests need to be especially thorough around preservation of unrelated XML content so we can be confident we’re not losing or reshaping metadata unintentionally outside the corrected field.

serverless/src/shared/applyDif10MetadataCorrections.js
htranho: Now added ProductLevelId

serverless/src/shared/__tests__/applyDif10MetadataCorrections.test.js
htranho: Done

serverless/src/shared/Dif10FieldsCorrection/applyProviderCorrection.js
htranho: As discussed, we remove only the keywords and let the validator do it's job.

PR #110, opened 2026-05-28
KMS-672: Safe DIF10 native metadata correction handling for prepared metadata-correction requests.

cgokey:
# Overview

### What is the feature?

KMS-672 adds safe DIF10 native metadata correction handling for prepared metadata-correction
requests.

Instead of rebuilding DIF10 through a generic object conversion, this branch applies targeted
DOM/XPath updates directly to the XML and then hands the corrected payload to a stubbed
CMR-writeback seam for follow-on work.

### What is the Solution?

This branch narrows `metadataCorrectionService` to one clear responsibility:

1. accept a fully prepared metadata-correction request
2. apply DIF10 keyword corrections safely
3. hand the corrected payload to the future CMR write component

Key changes:

- Replaces the old DIF10 whole-document rewrite flow with a shared XML DOM editor:
  - adds `XmlMetadataPathEditor`
  - adds `dif10DomEditor` as the DIF10-specific config layer
  - updates `applyDif10MetadataCorrections` to use targeted XML mutations instead of rebuilding
    XML from a parsed object
- Makes DIF10 matching path/value-based instead of index-based:
  - repeated DIF10 nodes are found by current keyword-path content
  - missing trailing hierarchy slots are treated as intentional blanks during matching
  - sparse path slots remain structurally significant
- Adds `writeCorrectedMetadataToCmr` as the internal writeback seam for the later external CMR
  persistence ticket.
- Preserves empty keyword path slots in shared path/cache helpers where that structure matters:
  - `splitKeywordPath`
  - `csvComparator`
  - `conceptForFullPathCacheBuilder`
  - `conceptForShortNameCacheBuilder`
- Adds a local DIF10 smoke script and checked-in fixtures that exercise:
  - the full supported DIF10 correction matrix
  - important editor special cases like empty-container pruning, empty-parent pruning, nested
    node creation, and scalar creation
- Documents the shared XML metadata path editor and the metadata-correction service contract in
  `README.md`.

- Metadata correction service request handling
- Shared DIF10 native metadata correction path
- Shared XML mutation utilities for future native formats
- Shared keyword-path parsing/comparison behavior where empty slots must be preserved
- Local smoke-test fixtures and scripts
- README architecture and contract documentation

# Testing

Run:

```bash
npx vite-node --config vite.config.js scripts/local/run_metadata_correction_dif10_smoke.mjs
```

The request fixture currently includes the full supported DIF10 correction matrix:

- `sciencekeywords`
- `locations`
- `chronounits`
- `platforms`
- `instruments`
- `projects`
- `providers`
- `rucontenttype`
- `idnnode`
- `isotopiccategory`
- `temporalresolutionrange`
- `verticalresolutionrange`
- `horizontalresolutionrange`
- `productlevelid`

The script prints a JSON summary like:

```json
{
  "collectionConceptId": "C1234567890-LOCAL",
  "nativeFormat": "DIF10",
  "requestedCorrections": 14,
  "appliedCorrections": 14
}
```

The important signals are:

- `nativeFormat` is `DIF10`
- `requestedCorrections` matches the fixture
- `appliedCorrections` matches the expected applied count
- `writeResult.stubbed` is `true`
- `writeResult.targetComponent` is `cmr-writeback`

The corrected XML is written to:

```bash
tmp/metadata_correction_smoke_output.xml
```

Useful spot checks:

- science keyword topic/term moved from the old aerosol path to the new oceans path
- chronostratigraphic unit `Epoch` changed from `HOLOCENE` to `PLEISTOCENE`
- platform `Short_Name` changed to `SPOT-4-UPDATED`
- platform `Long_Name` changed to `Systeme Observation de la Terre-4 Updated`
- instrument `GEOPHONES` is gone
- project `ALIENS` changed to `ALIENS-UPDATED`
- provider short name changed to `NZ/NZAI/ANZ-UPDATED`
- related URL content subtype changed from `OpenSearch` to `OGC WMS`
- `IDN_Node` no longer contains `CEOS`
- `ISO_Topic_Category` changed to `OCEANS`
- `Temporal_Resolution_Range` is gone
- `Vertical_Resolution_Range` changed to `50 meters`
- `Horizontal_Resolution_Range` changed to `100 meters`
- `Product_Level_Id` changed to `1A`

## For cloudwatch when we deploy to SIT

Look in the `metadataCorrectionService` log group
Current behavior in handler.js:
[metadata-correction] Received metadata correction request
includes the original metadataCorrectionRequest, so you get the requested correction objects there

[metadata-correction] Produced corrected metadata payload
now includes:
correctionCount
correctionsApplied
correctedMetadata

### Attachments

# Checklist


serverless/src/getConcepts/handler.js
```diff
@@ -231,12 +231,35 @@ export const getConcepts = async (event, context) => {
...
 
-      const csvResponse = await createCsvForScheme({
-        scheme: conceptScheme,
-        version,
-        versionName: keywordVersion,
-        versionCreationDate
-      })
+      let csvResponse
+      try {
+        const csvContent = await getCsvForScheme({
```
eudoroolivares2016:
```suggestion
        const csvContent = await createCsvForScheme({
```
This is def a nitpick but, since this is a mutation getCsvForSchema doesn't make as much sense to me as a name

serverless/src/getConcepts/handler.js
```diff
@@ -231,12 +231,35 @@ export const getConcepts = async (event, context) => {
...
+          versionCreationDate
+        })
+
+        csvResponse = {
+          statusCode: 200,
+          body: csvContent,
+          headers: {
+            ...defaultResponseHeaders,
+            'Content-Type': 'text/csv',
+            'Content-Disposition': `attachment; filename=${conceptScheme}.csv`
```
eudoroolivares2016, line 249: why doe we need the content disposition header here?

serverless/src/getConcepts/handler.js
```diff
@@ -231,12 +231,35 @@ export const getConcepts = async (event, context) => {
...
+          }
+        }
+      } catch (error) {
+        logger.error(`Error creating CSV for scheme=${conceptScheme}, error=${error.toString()}`)
```
eudoroolivares2016, line 253: Haven't looked yet but if that wasn't' a mutation then we'd just update this instead

serverless/src/shared/__tests__/downloadConcepts.test.js
```diff
@@ -77,7 +77,7 @@ describe('downloadConcepts', () => {
     })
 
     test('should download CSV content for a specific concept scheme', async () => {
-      const mockCsvContent = '"Category","Series_Entity","Short_Name","Long_Name","UUID"\n"EARTH OBSERVATION SATELLITES","TERRA","AM-1","Advanced Microwave Technology Satellite-1","d3c8e1f2-9b7a-4c5d-8e6f-1a2b3c4d5e6f"'
+      const mockCsvContent = '"Category","Class","Type","Short_Name","Long_Name","UUID"\n"EARTH REMOTE SENSING INSTRUMENTS","PASSIVE REMOTE SENSING","MICROWAVE RADIOMETERS","AM-1","Advanced Microwave Technology Satellite-1","d3c8e1f2-9b7a-4c5d-8e6f-1a2b3c4d5e6f"'
```
eudoroolivares2016, line 80: why did this one change?

serverless/src/shared/redis-path-store/helpers/buildKeywordObjectFromPath.js
```diff
@@ -0,0 +1,171 @@
+import { FULL_PATH_VALUE_FIELDS, SHORT_NAME_OBJECT_FIELDS } from './constants'
+import { isLookupShortNameScheme } from './isLookupShortNameScheme'
+import { joinKeywordPath } from './joinKeywordPath'
+import { normalizeKeywordScheme } from './normalizeKeywordScheme'
+import { splitKeywordPath } from './splitKeywordPath'
+import { trimKeywordPathSegment } from './trimKeywordPathSegment'
+
+function buildKeywordPathObjectFromSegments({
```
eudoroolivares2016:
```suggestion
const buildKeywordPathObjectFromSegments = ({
  scheme,
  segments
}) => {
  const slotFields = FULL_PATH_VALUE_FIELDS[normalizeKeywordScheme(scheme)]

  if (!Array.isArray(slotFields)) {
    return {}
  }

  const normalizedSegments = segments.map((segment) => trimKeywordPathSegment(segment))
  const paddedSegments = normalizedSegments.slice()

  while (paddedSegments.length < slotFields.length) {
    paddedSegments.push('')
  }

  return slotFields.reduce((keywordPathObject, fieldName, index) => ({
    ...keywordPathObject,
    [fieldName]: paddedSegments[index] || ''
  }), {})
}
```
just to make it consistent with the others but, nitpick

serverless/src/shared/writeCorrectedMetadataToCmr.js
```diff
@@ -0,0 +1,73 @@
...
+  return {
+    stubbed: true,
+    targetComponent: 'cmr-writeback',
+    collectionConceptId,
+    providerId,
+    nativeId,
+    nativeFormat,
+    correctionCount: Number(correctionCount || 0),
+    correctionsAppliedCount: Array.isArray(correctionsApplied) ? correctionsApplied.length : 0,
+    correctedMetadataBytes: Buffer.byteLength(serializedMetadata, 'utf8'),
```
eudoroolivares2016: Should we fallback on the serialize with a default value? if it fails here we don't need this field to return but, it'd fail the whole function

serverless/src/shared/redis-path-store/helpers/__tests__/flattenKeywordPathValue.test.js
```diff
@@ -0,0 +1,16 @@
...
+  test('preserves ordered scalar values from nested keyword content', () => {
+    expect(flattenKeywordPathValue({
+      Category: 'EARTH SCIENCE',
+      Topic: 'CRYOSPHERE',
+      VariableLevel1: 'SNOW/ICE'
+    })).toEqual(['EARTH SCIENCE', 'CRYOSPHERE', 'SNOW/ICE'])
+  })
+
+  test('returns empty arrays for nullish values and flattens nested arrays', () => {
+    expect(flattenKeywordPathValue(undefined)).toEqual([])
```
eudoroolivares2016, line 13:
```suggestion
  expect(flattenKeywordPathValue(undefined)).toEqual([])
    expect(flattenKeywordPathValue(null)).toEqual([])
    expect(flattenKeywordPathValue('')).toEqual([''])
```
easy coverage boost

serverless/src/shared/redis-path-store/helpers/parseCachedConceptResponse.js
```diff
@@ -0,0 +1,78 @@
+import { buildKeywordObjectFromPath } from './buildKeywordObjectFromPath'
+
+const attachSupplementalKeywordFields = ({
```
eudoroolivares2016, line 3: Do we need this since we only add the value to the cache if it exists already over in `serverless/src/shared/redis-path-store/helpers/createShortNameConceptResponseBody.js`

serverless/src/shared/getNumberOfCmrCollections.js
```diff
@@ -116,136 +117,34 @@ export const getNumberOfCmrCollections = async ({
...
+      scheme,
+      uuid,
+      prefLabel,
+      fullPath,
+      isLeaf
+    })
+
+    logger.debug('Mapped CMR scheme:', cmrScheme)
+
+    if (queryType === 'uuid') {
```
eudoroolivares2016, line 136: I might just ask to pair for this but, wouldn't we just only use the `uuid=<val?` api for all the coll schemes?

william-valencia (approved):
Overall looks pretty good. Did have some comments that were sent via slack.

serverless/src/getConcepts/handler.js
cgokey: Good catch. I renamed it to createCsvForScheme and updated the handler/tests accordingly. I agree that getCsvForScheme sounded more like a read than a CSV-generation step.

serverless/src/getConcepts/handler.js
cgokey: I think this was originally there to make the CSV behave like a downloadable file in browser/API clients and to provide a stable filename (${scheme}.csv). It isn’t new in this branch; it was part of the old CSV response shape and I preserved it when that response-building logic moved into the handler.

eudoroolivares2016:
Might not need to prop spread in `serverless/src/shared/invokeMetadataCorrectionDelegate.js` 


`import { applyDif10MetadataCorrections } from './applyDif10MetadataCorrections'
import { applyEcho10MetadataCorrections } from './applyEcho10MetadataCorrections'
import { applyIso19115MetadataCorrections } from './applyIso19115MetadataCorrections'
import { applyIsoSmapMetadataCorrections } from './applyIsoSmapMetadataCorrections'
import { applyUmmMetadataCorrections } from './applyUmmMetadataCorrections'

const normalizeKeywordObject = (keywordObject) => (
  keywordObject
  && typeof keywordObject === 'object'
  && !Array.isArray(keywordObject)
    ? keywordObject
    : {}
)

const normalizeCorrections = (corrections = []) => (
  Array.isArray(corrections)
    ? corrections.map((correction = {}) => ({
      ...correction,
      oldKeywordObject: normalizeKeywordObject(correction.oldKeywordObject),
      newKeywordObject: normalizeKeywordObject(correction.newKeywordObject)
    }))
    : []
)

/**
 * Routes correction plans to the appropriate native-format delegate.
 *
 * By the time this helper is called, the metadata-correction service has already:
 * - fetched the collection metadata
 * - validated supported keywords
 * - resolved invalid keywords into concrete replace/delete actions
 *
 * The remaining job is format-specific mutation. This helper provides that dispatch seam by
 * selecting the delegate that understands the collection's normalized native format and handing
 * it the correction plan plus any format-specific metadata payload it needs.
 *
 * That keeps the orchestration layer format-agnostic while allowing each delegate to own the
 * mechanics of mutating UMM, ISO19115, ISO SMAP, ECHO10, or DIF10 metadata.
 *
 * @param {object} params - Delegate parameters.
 * @param {'UMM'|'ISO19115'|'ISO_SMAP'|'ECHO10'|'DIF10'|'UNKNOWN'} params.nativeFormat - Normalized native format.
 * @returns {Promise<object>} Delegate result returned by the selected format-specific handler.
 * @throws {Error} If the native format does not have a registered delegate.
 */
export const invokeMetadataCorrectionDelegate = async ({
  nativeFormat,
  corrections,
  collectionConceptId
}) => {
  const normalizedParams = {
    corrections: normalizeCorrections(corrections),
    collectionConceptId
  }

  switch (nativeFormat) {
    case 'UMM':
      return applyUmmMetadataCorrections(normalizedParams)
    case 'ISO19115':
      return applyIso19115MetadataCorrections(normalizedParams)
    case 'ISO_SMAP':
      return applyIsoSmapMetadataCorrections(normalizedParams)
    case 'ECHO10':
      return applyEcho10MetadataCorrections(normalizedParams)
    case 'DIF10':
      return applyDif10MetadataCorrections(normalizedParams)
    default:
      throw new Error(`Unsupported native metadata format for delegate selection: ${nativeFormat}`)
  }
}

export default invokeMetadataCorrectionDelegate
`

eudoroolivares2016 (approved):

serverless/src/getConcepts/handler.js
cgokey: I don’t think we need to change it. This header was already part of the old CSV response behavior, and in the refactor I kept it in getConcepts/handler because it’s really an HTTP response concern, not part of CSV generation itself. createCsvForScheme now just builds the CSV content string, while the handler owns response headers/status, so this felt like the cleanest split while preserving the existing download behavior.

serverless/src/shared/__tests__/downloadConcepts.test.js
cgokey: This one wasn’t a behavior change in downloadConcepts. I updated the mocked CSV to match the actual current instruments CSV shape. The old fixture had a stale Series_Entity header, while the real published instruments CSV uses Category/Class/Type/Subtype/Short_Name/Long_Name/UUID. Since downloadConcepts just passes the CSV body through, the test would work either way; this was mainly fixture cleanup for accuracy.

serverless/src/shared/redis-path-store/helpers/buildKeywordObjectFromPath.js
cgokey: Yep, I cleaned this up. I ended up going a little further than just switching the local function style: there was already a shared buildKeywordPathObjectFromSegments helper, so I removed the duplicate local implementation and reused the shared one instead. That keeps the style consistent and trims a bit of duplicate logic too.

serverless/src/shared/writeCorrectedMetadataToCmr.js
cgokey: Good point, I added a fallback. correctedMetadataBytes is just telemetry, so it shouldn’t fail the whole function if serialization blows up. It now falls back to 0 in that case, and I added a focused test for a non-serializable payload.

serverless/src/shared/redis-path-store/helpers/__tests__/flattenKeywordPathValue.test.js
cgokey: Yep, I added those cases.

serverless/src/shared/redis-path-store/helpers/parseCachedConceptResponse.js
cgokey: Good point. It may be redundant for some freshly written short-name cache entries, but it’s still needed today because this parser also handles other cache shapes, including entries that only store fullPath and older/partial cached payloads. Worth a closer follow-up to see if we can tighten that contract and simplify it further.

serverless/src/shared/getNumberOfCmrCollections.js
cgokey: This is pre-existing behavior, not something introduced in this refactor. The current branching is still reflecting the older scheme-specific CMR query logic that was already here, and I moved the query construction behind createCmrCollectionQuery so at least that logic is centralized now. I agree the next step should be to keep migrating these older cases over to the newer query-builder API rather than continuing to reason about them inline in multiple places.   I just didn't want to change functionality quite yet in this refactor.

cgokey:
> Might not need to prop spread in `serverless/src/shared/invokeMetadataCorrectionDelegate.js`
> 
> `import { applyDif10MetadataCorrections } from './applyDif10MetadataCorrections' import { applyEcho10MetadataCorrections } from './applyEcho10MetadataCorrections' import { applyIso19115MetadataCorrections } from './applyIso19115MetadataCorrections' import { applyIsoSmapMetadataCorrections } from './applyIsoSmapMetadataCorrections' import { applyUmmMetadataCorrections } from './applyUmmMetadataCorrections'
> 
> const normalizeKeywordObject = (keywordObject) => ( keywordObject && typeof keywordObject === 'object' && !Array.isArray(keywordObject) ? keywordObject : {} )
> 
> const normalizeCorrections = (corrections = []) => ( Array.isArray(corrections) ? corrections.map((correction = {}) => ({ ...correction, oldKeywordObject: normalizeKeywordObject(correction.oldKeywordObject), newKeywordObject: normalizeKeywordObject(correction.newKeywordObject) })) : [] )
> 
> /**
> 
> * Routes correction plans to the appropriate native-format delegate.
> * By the time this helper is called, the metadata-correction service has already:
> * * fetched the collection metadata
> * * validated supported keywords
> * * resolved invalid keywords into concrete replace/delete actions
> * The remaining job is format-specific mutation. This helper provides that dispatch seam by
> * selecting the delegate that understands the collection's normalized native format and handing
> * it the correction plan plus any format-specific metadata payload it needs.
> * That keeps the orchestration layer format-agnostic while allowing each delegate to own the
> * mechanics of mutating UMM, ISO19115, ISO SMAP, ECHO10, or DIF10 metadata.
> * @param {object} params - Delegate parameters.
> * @param {'UMM'|'ISO19115'|'ISO_SMAP'|'ECHO10'|'DIF10'|'UNKNOWN'} params.nativeFormat - Normalized native format.
> * @returns {Promise} Delegate result returned by the selected format-specific handler.
> * @throws {Error} If the native format does not have a registered delegate.
>   */
>   export const invokeMetadataCorrectionDelegate = async ({
>   nativeFormat,
>   corrections,
>   collectionConceptId
>   }) => {
>   const normalizedParams = {
>   corrections: normalizeCorrections(corrections),
>   collectionConceptId
>   }
> switch (nativeFormat) { case 'UMM': return applyUmmMetadataCorrections(normalizedParams) case 'ISO19115': return applyIso19115MetadataCorrections(normalizedParams) case 'ISO_SMAP': return applyIsoSmapMetadataCorrections(normalizedParams) case 'ECHO10': return applyEcho10MetadataCorrections(normalizedParams) case 'DIF10': return applyDif10MetadataCorrections(normalizedParams) default: throw new Error(`Unsupported native metadata format for delegate selection: ${nativeFormat}`) } }
> 
> export default invokeMetadataCorrectionDelegate `

Good point. I tightened this up so we now normalize each correction into an explicit shared contract before delegation instead of passing through the whole object. That keeps all delegates receiving the same known fields (scheme, action, keywordConceptUuid, oldKeywordObject, newKeywordObject, ummPath, oldLongName, newLongName) while dropping unknown extras at this seam. I also added a test to lock that contract in.

PR #113, opened 2026-06-16
KMS-683: This PR implements the real metadata-correction writeback flow for collections and adds a synchronous operator endpoint t

cgokey:
# Overview

### What is the feature?

This PR implements the real metadata-correction writeback flow for collections and adds a synchronous operator endpoint that can run the correction pipeline for a single collection concept id and return a detailed result summary. It also adds the deployment/runtime wiring needed to safely gate live CMR writeback by token and provider allowlist.

### What is the Solution?

I pulled the core “fix one collection” logic out of the async SQS worker into a shared `runCollectionMetadataCorrection` helper so both the existing async service and a new sync API endpoint can use the same flow. That shared runner now handles UMM validation, invalid-keyword extraction, resolution into concrete corrections, native metadata delegate invocation, audit logging, and optional CMR ingest/writeback.

On the writeback side, I replaced the stub path with a real CMR PUT implementation using `CMR_WRITER_TOKEN` and rollout gating via `CMR_WRITEBACK_PROVIDERS`. I also threaded those env vars through CDK/Bamboo so the API and event-driven flows can write back when enabled. 

I added local helper scripts for calling the sync endpoint, fetching concepts, listing concept ids by native format, and running metadata-correction smoke tests. I also fixed a DIF10 matching issue in the shared XML editor so node matching is case-insensitive when locating an existing XML block, which was needed for real DIF10 payloads whose text casing differs from the normalized resolved keyword values.
We should probably do this for the JSON editor too.

- Metadata correction service flow
- New synchronous metadata correction API route: `PUT /metadata_correction/{collectionConceptId}`
- CMR writeback / ingest path
- CDK and Bamboo environment wiring for writeback config
- DIF10 XML matching behavior in the shared path editor
- Local smoke/debug scripts and README deployment/testing docs

# Testing

Provider for testing:
- AMD_KOPRI

Prerequisites:
- The branch is deployed to SIT.
- CMR_WRITER_TOKEN is configured in the deployed environment (it should have one already)
- CMR_WRITEBACK_PROVIDERS right now it is setup with AMD_KOPRI
- You have a valid MMT SIT bearer token for the protected KMS API.

1. Confirm there are DIF10 collections available in SIT and choose one from the AMD provider
   that has not already been corrected.

   Run:
   node scripts/local/list_collection_concept_ids_by_native_format.mjs sit dif10

   Verify the returned list includes at least one concept id ending in:
   -AMD_KOPRI

   Choose one AMD_KOPRI concept id from that list that has not already been fixed and export it:

   export CONCEPT_ID=<chosen AMD_KOPRI concept id>

2. Fetch the native metadata before running correction.

   Run:
   export CMR_AUTHORIZATION='Bearer <sit token>'
   ./scripts/local/fetch_concept.sh sit "$CONCEPT_ID" > before.txt

3. Run the synchronous metadata correction endpoint for the AMD collection.

   Run:
   export KMS_AUTHORIZATION='Bearer <sit token>'
   ./scripts/local/call_metadata_correction_sync_endpoint.sh sit "$CONCEPT_ID" | tee correction_response.json

4. Verify the correction response shows that keyword issues were found and that at least one
   correction was resolved and applied.

   Run:
   jq '.keywordValidationFailureCount, .resolvedCorrectionCount, .correctionResult.correctionCount, .writeResult.ingestResult' correction_response.json

   Expected:
   - keywordValidationFailureCount is greater than 0
   - resolvedCorrectionCount is greater than 0
   - correctionResult.correctionCount is greater than 0
   - writeResult.ingestResult.enabled is true
   - writeResult.ingestResult.updated is true

5. Fetch the native metadata again after correction.

   Run:
   ./scripts/local/fetch_concept.sh sit "$CONCEPT_ID" > after.txt

6. Diff the before and after native metadata payloads.

   Run:
   diff -u before.txt after.txt

   Verify the metadata changed in the expected keyword location. The diff should show the
   corrected DIF10 keyword content rather than an unchanged document.

7. Verify the audit lifecycle in the sync response.

   Inspect correction_response.json and confirm:
   - auditResults.pending is populated
   - auditResults.applied is populated

Notes:
- If the collection resolves no corrections, pick another AMD DIF10 collection and repeat the
  same flow.

### Attachments

# Checklist


serverless/src/shared/writeCorrectedMetadataToCmr.js
```diff
@@ -55,25 +221,135 @@ export const writeCorrectedMetadataToCmr = async ({
...
+      logger.info('[cmr-writeback] Skipping writeback because no writer token is configured', {
+        collectionConceptId,
+        providerId,
+        nativeId,
+        nativeFormat
+      })
+    }
+
+    return {
+      stubbed: false,
```
cgokey: drop this now

coverage-tmp/coverage-final.json
```diff
@@ -0,0 +1,3 @@
...,"1":27,"2":52,"3":29,"4":30,"5":12,"6":24,"7":16,"8":1,"9":1,"10":60,"11":3,"12":17,"13":35,"14":17,"15":60,"16":43,"17":63,"18":271,"19":58,"20":104,"21":30,"22":33,"23":57,"24":8,"25":1,"26":40,"27":30,"28":27,"29":1,"30":2,"31":12,"32":24,"33":18,"34":16,"35":12,"36":15,"37":18,"38":15,"39":12,"40":10,"41":3,"42":28,"43":22,"44":26,"45":13,"46":24,"47":5,"48":12,"49":24,"50":5,"51":8},"b":{"0":[172,27],"1":[29,1],"2":[1],"3":[8,4],"4":[4,3],"5":[1],"6":[60],"7":[63,1],"8":[58,19],"9":[32,1],"10":[57,57],"11":[3],"12":[3],"13":[8,4],"14":[4,4],"15":[8,4,4],"16":[1,3],"17":[2,1],"18":[2,28],"19":[0],"20":[22,5],"21":[15,1],"22":[3,12],"23":[2,1],"24":[1,11],"25":[1,10],"26":[10,1],"27":[2,9],"28":[9,1],"29":[2,1],"30":[3,3],"31":[1],"32":[5],"33":[2,26],"34":[23,3],"35":[2,1],"36":[3,2],"37":[2,20],"38":[14,6],"39":[13,1],"40":[14,13],"41":[1,13],"42":[13,1],"43":[1,12],"44":[1,12],"45":[11,2],"46":[6,1],"47":[5,1],"48":[4,1],"49":[12,1],"50":[1,11],"51":[2,9],"52":[1,1],"53":[8,1],"54":[2,22],"55":[13,9],"56":[1,7],"57":[8,1],"58":[1,7],"59":[5,1],"60":[1,4],"61":[2,2],"62":[1,1],"63":[1,1],"64":[8,1],"65":[8,5],"66":[2,6],"67":[1,1],"68":[5,1],"69":[1,4],"70":[2,2],"71":[1,1]}}
```
eudoroolivares2016: What is this for?

serverless/src/shared/getCmrWriterToken.js
```diff
@@ -0,0 +1,32 @@
...
+const getConfiguredWriterToken = () => String(process.env.CMR_WRITER_TOKEN || '').trim()
+
+/**
+ * Returns the bearer token used for CMR ingest/writeback requests.
+ *
+ * @returns {Promise<string>} Configured CMR bearer token.
+ * @throws {Error} If `CMR_WRITER_TOKEN` is empty or missing.
+ */
+export const getCmrWriterToken = async () => {
+  if (cachedWriterToken) {
```
eudoroolivares2016: can `cachedWriterToken` just be a const in this? not understanding why it needs to be global scope to this namespace

serverless/src/shared/__tests__/getCmrWriterToken.test.js
```diff
@@ -0,0 +1,60 @@
...
+describe('getCmrWriterToken', () => {
+  beforeEach(() => {
+    vi.resetModules()
+    delete process.env.CMR_WRITER_TOKEN
+  })
+
+  test('returns the configured environment token when present', async () => {
+    process.env.CMR_WRITER_TOKEN = 'writer-token'
+
+    const { getCmrWriterToken } = await import('../getCmrWriterToken')
```
eudoroolivares2016: Can we just import this once at the top its just the function?

serverless/src/shared/cmrPutRequest.js
```diff
@@ -0,0 +1,113 @@
...
+  } catch (error) {
+    const requestContext = {
+      method: 'PUT',
+      endpoint,
+      path,
+      fullUrl,
+      bodyLength: typeof body === 'string' ? body.length : undefined
+    }
+
+    logger.error('[cmr-put] CMR fetch failed', {
```
eudoroolivares2016:
```suggestion
    logger.error('[cmr-put] CMR write failed', {
```

serverless/src/shared/runCollectionMetadataCorrection.js
```diff
@@ -0,0 +1,384 @@
...
+
+/**
+ * True when the current native format is supported by the active runtime mode.
+ *
+ * DIF10 is supported everywhere. UMM is supported only for local smoke/development mode.
+ *
+ * @param {string} nativeFormat Normalized native format label.
+ * @returns {boolean} `true` when the runner should process this format.
+ */
+const isSupportedNativeFormat = (nativeFormat) => (
```
eudoroolivares2016: Aren't we doing this in another namespace on the case statement I guess this is fine as a temp stub

eudoroolivares2016 (approved):

htranho (approved):

coverage-tmp/coverage-final.json
cgokey: Removed

serverless/src/shared/getCmrWriterToken.js
cgokey: I agree and simplified this. Since CMR_WRITER_TOKEN is now a plain env var, the module-level cache wasn’t buying us anything meaningful, so I removed it and just read/trim the env var directly on each call.

serverless/src/shared/__tests__/getCmrWriterToken.test.js
cgokey: Agree, that dynamic import was left over from when the helper had module-level cached state. Since it now just reads the env var directly, I simplified the test to a normal top-level import.

serverless/src/shared/cmrPutRequest.js
cgokey: Agree, I'll update this.

serverless/src/shared/runCollectionMetadataCorrection.js
cgokey: Good call. We were duplicating the format support between runCollectionMetadataCorrection and the delegate code. I refactored that so the source of truth now lives in invokeMetadataCorrectionDelegate via a shared isMetadataCorrectionDelegateSupported() helper, and the runner just asks that helper before proceeding. That keeps the temporary DIF9/UMM gating behavior, but removes the drift risk from having the support list in two places.

PR #114, opened 2026-06-29
KMS-669: Add monitoring and metrics for event processing

cgokey:
# Overview

### What is the feature?

This PR adds CloudWatch monitoring for metadata-correction consumer processing.

This gives us visibility into:
- how many async correction events are consumed
- how many finish successfully or fail
- how many invalid keywords are found and resolved
- how many corrections are applied to metadata
- how many corrected records are actually written back to CMR

It also adds equivalent metrics for the manual single-collection sync path so both flows report into the same `CMR/KeywordSync` namespace.

### What is the Solution?

This branch adds shared consumer metric emission and wires it into both metadata-correction execution paths.

Key changes:
- Adds shared metric definitions and emission logic in:
  - `serverless/src/shared/emitConsumerMetrics.js`
  - `serverless/src/shared/emitConsumerMetricsSafely.js`
- Emits consumer lifecycle metrics from the async metadata-correction consumer:
  - `EventsConsumed`
  - `EventsProcessed`
  - `EventProcessingFailures`
- Emits correction-stage and writeback-stage metrics from the shared collection runner:
  - `InvalidKeywordCount`
  - `KeywordsResolved`
  - `CorrectionsAppliedToMetadata`
  - `CorrectionsWrittenToCMR`
  - `RecordsUpdatedFromEvent`
  - `RecordsUpdatedFromManual`
- Adds CloudWatch `PutMetricData` permissions for the metadata correction setup in CDK.

- Metadata correction consumer
- Shared correction runner
- Shared metrics helpers
- CDK IAM setup for metadata correction

# Testing

# KMS-669 AWS Test Session

Script used:
- `scripts/local/find_collection_with_positive_manual_sync_result.mjs`

```bash
# Set the KMS token used by the manual correction endpoint.
export KMS_AUTHORIZATION='Bearer <kms token>'

# Run the manual correction finder until one AMD_KOPRI DIF10 collection returns a positive result.
ALLOW_MUTATING_CORRECTION_ENDPOINT=true MAX_CONCEPTS=10 \
node scripts/local/find_collection_with_positive_manual_sync_result.mjs sit dif10 AMD_KOPRI \
  | tee /tmp/kms-669-correction-response.json

# Inspect the key response fields from the matching correction run.
jq '{ collectionConceptId, matchedField, matchedValue, outcome: .response.outcome, keywordValidationFailureCount: .response.keywordValidationFailureCount, resolvedCorrectionCount: .response.resolvedCorrectionCount, correctionsAppliedCount: .response.correctionResult.correctionsAppliedCount, writeUpdated: .response.writeResult.ingestResult.updated }' \
  /tmp/kms-669-correction-response.json
```

In AWS, open `CloudWatch` -> `Metrics` -> `All metrics` -> `Custom namespaces` -> `CMR/KeywordSync`. Use `Sum`, set the period to `1 minute`, and check the last few minutes right after running the script. For a good manual test run, look for movement on `InvalidKeywordCount`, `KeywordsResolved`, and `CorrectionsAppliedToMetadata`. If writeback was enabled and succeeded, you should also see `CorrectionsWrittenToCMR` and `RecordsUpdatedFromManual`.

### Attachments

# Checklist

## Summary by CodeRabbit

* **New Features**
  * Added consumer keyword-sync metric reporting for metadata correction runs, including consumed/processed and processing-failure metrics.
  * Updated local fixtures and mock server to support native metadata content type overrides.
  * Added local smoke tests for async and manual sync flows, plus a helper script to find collections with positive manual-sync results.
* **Bug Fixes**
  * Improved batch handling so all records are attempted; metric emission failures are logged but no longer block processing.
* **Tests**
  * Expanded unit and integration coverage for metric emission and correction scenarios, including error and no-op paths.

### DIFF
--- a/cdk/app/lib/helper/MetadataCorrectionSetup.ts
+++ b/cdk/app/lib/helper/MetadataCorrectionSetup.ts
@@ -161,7 +161,12 @@ export class MetadataCorrectionSetup extends Construct {
     this.metadataCorrectionRequestsQueue.grantConsumeMessages(this.metadataCorrectionServiceLambda)
     this.metadataCorrectionServiceLambda.addToRolePolicy(new iam.PolicyStatement({
       actions: ['cloudwatch:PutMetricData'],
-      resources: ['*']
+      resources: ['*'],
+      conditions: {
+        StringEquals: {
+          'cloudwatch:namespace': 'CMR/KeywordSync'
+        }
+      }
     }))
 
     this.metadataCorrectionRequestsTopicArnOutput = new cdk.CfnOutput(this, 'MetadataCorrectionRequestsTopicArn', {
--- a/cdk/app/lib/helper/MetadataCorrectionSetup.ts
+++ b/cdk/app/lib/helper/MetadataCorrectionSetup.ts
@@ -2,6 +2,7 @@ import * as path from 'path'
 
 import * as cdk from 'aws-cdk-lib'
 import * as ec2 from 'aws-cdk-lib/aws-ec2'
+import * as iam from 'aws-cdk-lib/aws-iam'
 import * as eventsources from 'aws-cdk-lib/aws-lambda-event-sources'
 import { NodejsFunction } from 'aws-cdk-lib/aws-lambda-nodejs'
 import * as sns from 'aws-cdk-lib/aws-sns'
@@ -158,6 +159,10 @@ export class MetadataCorrectionSetup extends Construct {
     ))
 
     this.metadataCorrectionRequestsQueue.grantConsumeMessages(this.metadataCorrectionServiceLambda)
+    this.metadataCorrectionServiceLambda.addToRolePolicy(new iam.PolicyStatement({
+      actions: ['cloudwatch:PutMetricData'],
+      resources: ['*']
+    }))
 
     this.metadataCorrectionRequestsTopicArnOutput = new cdk.CfnOutput(this, 'MetadataCorrectionRequestsTopicArn', {
       description: 'SNS topic ARN for metadata correction request publishing',
--- a/scripts/local/find_collection_with_positive_manual_sync_result.mjs
+++ b/scripts/local/find_collection_with_positive_manual_sync_result.mjs
@@ -37,6 +37,7 @@ const showUsageAndExit = (exitCode) => {
     '  MIN_VALUE            Minimum numeric value to count as a match. Defaults to 1.',
     '  MAX_CONCEPTS         Optional cap on how many concept ids to try.',
     '  REQUEST_DELAY_MS     Optional delay between endpoint calls. Defaults to 0.',
+    '  STOP_ON_ERROR        When true, throw on the first non-2xx response instead of skipping it.',
     '  KMS_BASE_URL         Optional KMS base URL override.',
     '',
     'Safety:',
@@ -206,6 +207,17 @@ const getFieldValue = (responseBody, fieldPath) => fieldPath
   ), responseBody)
 
 /**
+ * Compacts an error response body into a one-line snippet for stderr logging.
+ *
+ * @param {string} bodyText Raw HTTP response body.
+ * @returns {string} Trimmed one-line body preview.
+ */
+const formatErrorBodySnippet = (bodyText) => String(bodyText || '')
+  .replace(/\s+/g, ' ')
+  .trim()
+  .slice(0, 500)
+
+/**
  * Calls the synchronous metadata-correction endpoint for one concept id.
  *
  * @param {object} params Request inputs.
@@ -303,6 +315,7 @@ const buildMatchSummary = ({
  * @param {string} params.resultField Dot-path field to inspect on the JSON response.
  * @param {number} params.minValue Minimum numeric value that counts as a match.
  * @param {number} params.requestDelayMs Delay between requests.
+ * @param {boolean} params.stopOnError When true, throw on the first non-2xx response.
  * @param {number} [params.inspectedCount=0] Number of requests already attempted.
  * @returns {Promise<object>} Match summary or a no-match summary when exhausted.
  */
@@ -313,6 +326,7 @@ const findFirstMatchingCollection = async ({
   resultField,
   minValue,
   requestDelayMs,
+  stopOnError,
   inspectedCount = 0
 }) => {
   if (candidateConceptIds.length === 0) {
@@ -340,10 +354,19 @@ const findFirstMatchingCollection = async ({
   })
 
   if (!result.ok) {
+    const errorBodySnippet = formatErrorBodySnippet(result.bodyText)
+
     process.stderr.write(
       `[find-positive-manual-sync-result] Skipping ${collectionConceptId}; `
-      + `endpoint returned ${result.status}\n`
+      + `endpoint returned ${result.status}${errorBodySnippet ? ` body=${errorBodySnippet}` : ''}\n`
     )
+
+    if (stopOnError) {
+      throw new Error(
+        `Manual sync endpoint failed for ${collectionConceptId} `
+        + `with status ${result.status}. ${errorBodySnippet || 'No response body returned.'}`
+      )
+    }
 
     if (requestDelayMs > 0) {
       await sleep(requestDelayMs)
@@ -356,6 +379,7 @@ const findFirstMatchingCollection = async ({
       resultField,
       minValue,
       requestDelayMs,
+      stopOnError,
       inspectedCount: nextInspectedCount
     })
   }
@@ -387,6 +411,7 @@ const findFirstMatchingCollection = async ({
     resultField,
     minValue,
     requestDelayMs,
+    stopOnError,
     inspectedCount: nextInspectedCount
   })
 }
@@ -420,6 +445,7 @@ const main = async () => {
   const resultField = String(process.env.RESULT_FIELD || 'resolvedCorrectionCount').trim()
   const minValue = parsePositiveInteger(process.env.MIN_VALUE, 1, 'MIN_VALUE')
   const requestDelayMs = parseNonNegativeInteger(process.env.REQUEST_DELAY_MS, 0, 'REQUEST_DELAY_MS')
+  const stopOnError = String(process.env.STOP_ON_ERROR || '').toLowerCase() === 'true'
   const maxConcepts = process.env.MAX_CONCEPTS
     ? parsePositiveInteger(process.env.MAX_CONCEPTS, 1, 'MAX_CONCEPTS')
     : undefined
@@ -455,7 +481,8 @@ const main = async () => {
     authorizationValue,
     resultField,
     minValue,
-    requestDelayMs
+    requestDelayMs,
+    stopOnError
   })
 
   process.stdout.write(`${JSON.stringify(summary, null, 2)}\n`)
--- a/scripts/local/find_collection_with_positive_manual_sync_result.mjs
+++ b/scripts/local/find_collection_with_positive_manual_sync_result.mjs
@@ -1,0 +1,464 @@
+#!/usr/bin/env node
+
+import { spawn } from 'node:child_process'
+
+const ENVIRONMENT_BASE_URLS = {
+  sit: 'https://cmr.sit.earthdata.nasa.gov/kms',
+  uat: 'https://cmr.uat.earthdata.nasa.gov/kms',
+  prod: 'https://cmr.earthdata.nasa.gov/kms'
+}
+
+/**
+ * Prints usage guidance and exits.
+ *
+ * This helper calls the real synchronous metadata-correction endpoint, so it can
+ * mutate CMR metadata when writer-token rollout is enabled in the deployed env.
+ *
+ * @param {number} exitCode Process exit code.
+ * @returns {never} Always exits the process.
+ */
+const showUsageAndExit = (exitCode) => {
+  const message = [
+    'Usage:',
+    '  KMS_AUTHORIZATION=\'Bearer <token>\' \\',
+    '  ALLOW_MUTATING_CORRECTION_ENDPOINT=true \\',
+    '  node scripts/local/find_collection_with_positive_manual_sync_result.mjs <sit|uat|prod> [format] [providerId]',
+    '',
+    'Examples:',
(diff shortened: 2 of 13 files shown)


cdk/app/lib/helper/MetadataCorrectionSetup.ts
```diff
@@ -158,6 +159,10 @@ export class MetadataCorrectionSetup extends Construct {
     ))
 
     this.metadataCorrectionRequestsQueue.grantConsumeMessages(this.metadataCorrectionServiceLambda)
+    this.metadataCorrectionServiceLambda.addToRolePolicy(new iam.PolicyStatement({
+      actions: ['cloudwatch:PutMetricData'],
+      resources: ['*']
+    }))
```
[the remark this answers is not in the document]
cgokey: This grant still needs to live here because metadataCorrectionServiceLambda is created with its own execution role in MetadataCorrectionSetup and does not inherit the shared IamSetup.lambdaRole. I tightened it by scoping cloudwatch:PutMetricData to the CMR/KeywordSync namespace, so we keep the permission the lambda needs while making it more least-privilege.

scripts/local/run_metadata_correction_consumer_metrics_async_smoke.mjs
```diff
@@ -0,0 +1,676 @@
...
+const fixturePath = path.resolve(
+  rootDir,
+  'scripts/local/fixtures/metadata_correction_smoke.full_path.example.json'
+)
+const outputDir = path.resolve(rootDir, 'tmp/metadata-correction-consumer-metrics-async-smoke')
+const outputPath = path.resolve(outputDir, 'result.json')
+const mockCmrPort = Number(process.env.MOCK_CMR_PORT || 3020)
+const cmrBaseUrl = process.env.CMR_BASE_URL || `http://127.0.0.1:${mockCmrPort}`
+const cloudWatchEndpoint = process.env.AWS_ENDPOINT_URL || 'http://127.0.0.1:4566'
+const startMockServer = String(process.env.START_MOCK_CMR || 'true').toLowerCase() !== 'false'
```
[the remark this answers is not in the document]
cgokey: I simplified both consumer metric smoke scripts so they are now explicitly local-only. They no longer honor external CMR_BASE_URL or AWS_ENDPOINT_URL overrides, and always target the local mock CMR plus local LocalStack CloudWatch. That keeps the smoke simpler and avoids the risk of accidentally health-checking or writing against the wrong backend.

scripts/local/run_metadata_correction_consumer_metrics_manual_sync_smoke.mjs
```diff
@@ -0,0 +1,687 @@
...
+const fixturePath = path.resolve(
+  rootDir,
+  'scripts/local/fixtures/metadata_correction_smoke.full_path.example.json'
+)
+const outputDir = path.resolve(rootDir, 'tmp/metadata-correction-consumer-metrics-manual-sync-smoke')
+const outputPath = path.resolve(outputDir, 'result.json')
+const mockCmrPort = Number(process.env.MOCK_CMR_PORT || 3020)
+const cmrBaseUrl = process.env.CMR_BASE_URL || `http://127.0.0.1:${mockCmrPort}`
+const cloudWatchEndpoint = process.env.AWS_ENDPOINT_URL || 'http://127.0.0.1:4566'
+const startMockServer = String(process.env.START_MOCK_CMR || 'true').toLowerCase() !== 'false'
```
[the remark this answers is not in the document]
cgokey: I simplified both consumer metric smoke scripts so they are now explicitly local-only. They no longer honor external CMR_BASE_URL or AWS_ENDPOINT_URL overrides, and always target the local mock CMR plus local LocalStack CloudWatch. That keeps the smoke simpler and avoids the risk of accidentally health-checking or writing against the wrong backend.

scripts/local/find_collection_with_positive_manual_sync_result.mjs
```diff
@@ -0,0 +1,491 @@
...
+  }
+
+  const { environment, baseUrl } = resolveEnvironment(environmentArg)
+  const resultField = String(process.env.RESULT_FIELD || 'resolvedCorrectionCount').trim()
+  const minValue = parsePositiveInteger(process.env.MIN_VALUE, 1, 'MIN_VALUE')
+  const requestDelayMs = parseNonNegativeInteger(process.env.REQUEST_DELAY_MS, 0, 'REQUEST_DELAY_MS')
+  const stopOnError = String(process.env.STOP_ON_ERROR || '').toLowerCase() === 'true'
+  const maxConcepts = process.env.MAX_CONCEPTS
+    ? parsePositiveInteger(process.env.MAX_CONCEPTS, 1, 'MAX_CONCEPTS')
+    : undefined
```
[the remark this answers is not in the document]
cgokey: Even though this is a local helper, it can issue mutating PUT calls against shared environments, so I added a small blast-radius guard for both uat and prod. Those runs now require either an explicit MAX_CONCEPTS limit or a providerId filter before the script will proceed.

scripts/local/find_collection_with_positive_manual_sync_result.mjs
```diff
@@ -0,0 +1,491 @@
...
+    })
+  }
+
+  const resultValue = Number(getFieldValue(result.responseBody, resultField) || 0)
+
+  process.stderr.write(
+    `[find-positive-manual-sync-result] ${collectionConceptId} -> ${resultField}=${resultValue}\n`
+  )
+
+  if (resultValue >= minValue) {
```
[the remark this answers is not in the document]
cgokey: I tightened the script so RESULT_FIELD must resolve to a real numeric value in the response. If the field is missing or non-numeric, the script now fails immediately instead of defaulting to 0 and continuing to issue mutating requests across the candidate list.

serverless/src/shared/emitConsumerMetricsSafely.js
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Emits consumer metrics without failing the caller when CloudWatch is unavailable.
+ *
+ * @param {Object} params Safe emission details.
+ * @param {Array<{metricName: string, value: number}>} params.metrics Metrics to emit.
+ * @param {string} params.logMessage Error log message to use when metric emission fails.
+ * @param {Object} [params.logContext={}] Structured log context for failed emission.
+ * @returns {Promise<void>}
+ */
+export const emitConsumerMetricsSafely = async ({
```
eudoroolivares2016, line 18: Why are we doing this?

scripts/local/fixtures/metadata_correction_smoke.full_path.example.json
```diff
@@ -132,6 +132,7 @@
         "nativeId": "local-smoke-collection-1",
         "providerId": "LOCAL",
         "format": "application/vnd.nasa.cmr.umm+json",
+        "nativeMetadataContentType": "application/vnd.nasa.cmr.umm+json;version=1.16.2; charset=utf-8",
```
eudoroolivares2016: This field isn't part of umm-c though on the meta section I know this is in the metadataspecification but, its being pulled here into where `meta` is?

serverless/src/shared/runCollectionMetadataCorrection.js
```diff
@@ -153,6 +155,86 @@ const determineNoOpOutcome = (keywordValidationFailures) => (
...
+}) => {
+  if (source === 'metadataCorrectionApi') {
+    return CONSUMER_METRIC_NAMES.RECORDS_UPDATED_FROM_MANUAL
+  }
+
+  if (typeof keywordEvent?.uuid === 'string' && keywordEvent.uuid.trim().length > 0) {
+    return CONSUMER_METRIC_NAMES.RECORDS_UPDATED_FROM_EVENT
+  }
+
+  return undefined
```
eudoroolivares2016: Maybe log here or even throw an error shouldn't get here anyway

serverless/src/shared/runCollectionMetadataCorrection.js
```diff
@@ -153,6 +155,86 @@ const determineNoOpOutcome = (keywordValidationFailures) => (
...
+ */
+const getRecordUpdateMetricName = ({
+  keywordEvent,
+  source
+}) => {
+  if (source === 'metadataCorrectionApi') {
+    return CONSUMER_METRIC_NAMES.RECORDS_UPDATED_FROM_MANUAL
+  }
+
+  if (typeof keywordEvent?.uuid === 'string' && keywordEvent.uuid.trim().length > 0) {
```
eudoroolivares2016: Can we just pull off the event here `"action": "UPDATED",` and then use MANUAL and DELETE?

serverless/src/metadataCorrectionService/handler.js
```diff
@@ -15,6 +17,17 @@ export const metadataCorrectionService = async (event) => {
   const records = event?.Records || []
 
   await Promise.all(records.map(async (record) => {
+    await emitConsumerMetricsSafely({
```
eudoroolivares2016: does this need to run in the loop or can we just keep a counter and then push once at the end with the value?

serverless/src/metadataCorrectionService/handler.js
```diff
@@ -15,6 +17,17 @@ export const metadataCorrectionService = async (event) => {
...
+      metrics: [{
+        metricName: CONSUMER_METRIC_NAMES.EVENTS_CONSUMED,
+        value: 1
+      }],
+      logMessage: '[metadata-correction] Failed to emit processing metrics',
```
eudoroolivares2016: Kind of confused about this log message here

serverless/src/shared/emitConsumerMetrics.js
```diff
@@ -0,0 +1,153 @@
...
+    body: buildCloudWatchQueryRequestBody({
+      metrics
+    })
+  })
+
+  if (!response.ok) {
+    const responseBody = await response.text()
+
+    throw new Error(
+      'Failed to emit consumer keyword sync metrics to LocalStack. '
```
eudoroolivares2016:
```suggestion
      'Failed to emit consumer keyword sync metrics. '
```

serverless/src/shared/emitConsumerMetrics.js
```diff
@@ -0,0 +1,153 @@
...
+      metrics
+    })
+  } else {
+    await cloudWatchClient.send(new PutMetricDataCommand({
+      Namespace: CONSUMER_METRIC_NAMESPACE,
+      MetricData: metricData
+    }))
+  }
+
+  logger.info(
```
eudoroolivares2016: Do we need logger if we are writing to a different metric feels like noise think we at least make this a logger.debugger?

serverless/src/shared/emitConsumerMetricsSafely.js
cgokey: I pulled this into a shared helper because we now emit consumer metrics from both the async worker and the shared collection-correction flow, and in both places the behavior should be the same: metric emission is best-effort and must not fail the actual correction run. This helper keeps that “try to emit, log context on failure, continue processing” behavior in one place instead of duplicating the same try/catch block at each call site.  I've updated the comments to better explain this.

scripts/local/fixtures/metadata_correction_smoke.full_path.example.json
cgokey: You’re right that this field is not part of real UMM-C. The actual code depends on the real .native response Content-Type header from CMR, not on a UMM payload field. This fixture value is only mock-server metadata so the local fake CMR knows how to return the same header shape as production. I renamed it to nativeResponseContentType to make that clearer.

serverless/src/shared/runCollectionMetadataCorrection.js
cgokey: Makes sense. I don’t want to throw adding debug log on that path so it’s visible if we ever hit it unexpectedly.

serverless/src/shared/runCollectionMetadataCorrection.js
cgokey: I’d prefer to keep this source-based rather than action-based. The manual endpoint has no keyword event, so it needs its own bucket, and the async path is still the same event-driven flow whether the incoming eventType was UPDATED or DELETED. In other words, this metric is answering “where did this successful writeback come from?” not “what kind of keyword event triggered it?” If we want UPDATED vs DELETED visibility later, I think that should be a separate metric split rather than replacing the manual/event distinction.

serverless/src/metadataCorrectionService/handler.js
cgokey: Good call. I updated this to tabulate the batch outcomes and emit the processing metrics once at the end of the handler rather than per record. EventsConsumed is now records.length, and EventsProcessed / EventProcessingFailures are emitted from the aggregated success/failure counts for the batch.

serverless/src/metadataCorrectionService/handler.js
cgokey: Agreed, that message was too generic. I renamed it to Failed to emit async batch processing metrics so it’s clearer this is the async consumer’s batch-level metric emission path, not the shared per-collection correction metrics.

serverless/src/shared/emitConsumerMetrics.js
cgokey: Agree and changed

serverless/src/shared/emitConsumerMetrics.js
cgokey: Agree switching to debug

serverless/src/shared/runCollectionMetadataCorrection.js
eudoroolivares2016: Manual endpoint has one I thought since its over with https://cmr.sit.earthdata.nasa.gov/kms/metadata_correction_audit?action=MANUAL?

eudoroolivares2016 (approved):

serverless/src/shared/runCollectionMetadataCorrection.js
cgokey: I see, yeah agree.  I updated this to keep the existing metric names and just use action to choose the bucket: MANUAL emits RecordsUpdatedFromManual, and UPDATED/DELETED emit RecordsUpdatedFromEvent.

htranho (approved):

PR #115, opened 2026-06-30
KMS-686: Update Keywords in ISO19115 native records

htranho:
# Overview

### What is the feature?

Update Keywords in ISO19115 (ISO-MENDS) native records

### What is the Solution?

Update keywords in following schemes as they appear in ISO-MENDS records, according to correction requests:

sciencekeywords
locations
platforms
instruments
projects
providers
isotopiccategory
productlevelid

Consumer - Update Keywords in ISO-MENDS native records

# Testing

Run:

```bash
npx vite-node --config vite.config.js scripts/local/run_metadata_correction_iso-19115_smoke.mjs
```

The request fixture currently includes the supported UMM-C correction matrix:

- `sciencekeywords`
- `locations`
- `platforms`
- `instruments`
- `projects`
- `providers`
- `isotopiccategory`
- `productlevelid`

The script prints a JSON summary like:

```json
[metadata-correction-smoke] Applied corrections locally
{
  "fixturePath": "/Users/htranho/repos/kms/scripts/local/fixtures/metadata_correction_service.iso-19115.corrections.json",
  "collectionConceptId": "C1234567222-LOCAL",
  "nativeFormat": "ISO19115",
  "requestedCorrections": 10,
  "appliedCorrections": 10,
  "outputPath": "/Users/htranho/repos/kms/tmp/metadata_correction_service_iso-19115_smoke_output.xml",
  "writeResult": {
    "targetComponent": "cmr-writeback",
    "collectionConceptId": "C1234567222-LOCAL",
    "providerId": null,
    "nativeId": null,
    "nativeFormat": "ISO19115",
    "correctionCount": 10,
    "correctionsAppliedCount": 10,
    "correctedMetadataBytes": 16144,
    "source": "local-smoke",
    "ingestResult": {
      "enabled": false,
      "ingested": false,
      "updated": false
    }
  }
}

```

The important signals are:

- `nativeFormat` is `ISO19115`
- `requestedCorrections` matches the fixture
- `appliedCorrections` matches the expected applied count
- `writeResult.targetComponent` is `cmr-writeback`

The corrected XML is written to:

```bash
tmp/metadata_correction_service_iso-19115_smoke_output.xml
```

See scripts/local/fixtures/metadata_correction_service.iso-19115.corrections.json for requested corrections and compare them with result in tmp/metadata_correction_service_iso-19115_smoke_output.xml

# Checklist

## Summary by CodeRabbit

* **New Features**
  * Apply ISO-19115 metadata corrections across science keywords, locations, platforms, instruments, projects, providers, topic categories, and processing levels.
  * Added native-format aliases `iso19115` and `isosmap` for command-line resolution.
  * Introduced an ISO-19115 XML editor and scheme-based keyword editors to update and delete targeted nodes.
* **Bug Fixes**
  * Corrections now perform real replace/delete updates with proper cleanup (including removal of emptied containers) and synchronized processing-level edits.
* **Tests**
  * Expanded ISO-19115 coverage with new fixtures, a new smoke-test workflow, and expanded editor behavior assertions.
* **Chores**
  * Improved shared XML utilities and namespace/XPath-based editing helpers.

### DIFF
--- a/serverless/src/shared/Iso19115DomEditor.js
+++ b/serverless/src/shared/Iso19115DomEditor.js
@@ -4,14 +4,22 @@ import { FULL_PATH_VALUE_FIELDS } from './redis-path-store/helpers/constants'
 /**
  * Helper factory function to create a block editor configuration.
  * Maps a correction to an update operation within the editor instance.
+ * @param {Object} config - Configuration object defining XPath and transformation logic.
+ * @returns {Function} Function to apply the update.
  */
 const blockScheme = (config) => (editor, correction) => editor.updateBlockNode(correction, config)
-
+/**
+ * Helper factory function to create a leaf editor configuration.
+ * @param {Object} config - Configuration object for updating single nodes.
+ * @returns {Function} Function to apply the update.
+ */
 const leafScheme = (config) => (editor, correction) => editor.updateLeafNode(correction, config)
-
 /**
  * Factory to generate standardized keyword block editors for structures
  * like Platforms and Instruments.
+ * * @param {string} type - The 'codeListValue' for the MD_KeywordTypeCode.
+ * @param {Object} options - Configuration for field parsing and value generation.
+ * @returns {Function} Configured block editor function.
  */
 const createKeywordBlock = (type, { fieldKeys, matchKeys, getValue }) => blockScheme({
   nodeXPath: `//gmd:descriptiveKeywords/gmd:MD_Keywords[gmd:type/gmd:MD_KeywordTypeCode/@codeListValue = '${type}']`,
@@ -19,6 +27,10 @@ const createKeywordBlock = (type, { fieldKeys, matchKeys, getValue }) => blockSc
     fieldPaths: ['gmx:Anchor', 'gco:CharacterString'],
     valueKeys: fieldKeys,
     matchKeys,
+    /**
+     * Extracts values from the XML node and maps them to field keys.
+     * Handles hierarchical strings (l1 > l2).
+     */
     getNodeValueObject: ({ node, editor }) => {
       const anchorNode = editor.selectNodes('./gmx:Anchor', node)[0]
       const charStringNode = editor.selectNodes('./gco:CharacterString', node)[0]
@@ -36,16 +48,20 @@ const createKeywordBlock = (type, { fieldKeys, matchKeys, getValue }) => blockSc
   },
   replace: [
     {
+      // Dynamically select target element (Anchor vs CharacterString) based on existing content
       fieldPath: ({ node, editor }) => (editor.selectNodes('./gmx:Anchor', node).length > 0 ? 'gmx:Anchor' : 'gco:CharacterString'),
       source: {
         type: 'computed',
-        // Use the custom getValue if provided, else fall back to default
+        // Allows custom formatting for keyword strings; defaults to joining keys with ' > '
         getValue: getValue || (({ correction }) => fieldKeys.map((k) => correction.newKeywordObject[k]).filter(Boolean).join(' > '))
       }
     }
   ]
 })
-
+/**
+ * Creates an editor for ISO Topic Category nodes.
+ * Updates both the text content and the @codeListValue attribute.
+ */
 const createIsoTopicCategoryEditor = () => leafScheme({
   nodeXPath: '//gmd:identificationInfo/gmd:MD_DataIdentification/gmd:topicCategory',
   find: {
@@ -70,7 +86,10 @@ const createIsoTopicCategoryEditor = () => leafScheme({
     }
   ]
 })
-
+/**
+ * Creates an editor for Processing Level Identifiers.
+ * Manages deletion of old paths and insertion/updates into specific XML locations.
+ */
 const createProductLevelIdEditor = () => leafScheme({
   nodeXPath: '//gmd:processingLevel/gmd:MD_Identifier[gmd:codeSpace/gco:CharacterString="gov.nasa.esdis.umm.processinglevelid"]',
   find: {
@@ -85,7 +104,6 @@ const createProductLevelIdEditor = () => leafScheme({
   ],
   replace: [
     {
-      // Since matchingNode is now MD_Identifier, this path correctly selects the child
       fieldPath: 'gmd:code/gco:CharacterString',
       source: {
         type: 'computed',
@@ -93,6 +111,7 @@ const createProductLevelIdEditor = () => leafScheme({
       }
     },
     {
+      // Target specific secondary locations for synchronization
       fieldPath: '//gmd:contentInfo/gmd:MD_ImageDescription/gmd:processingLevelCode/gmd:MD_Identifier[gmd:codeSpace/gco:CharacterString="gov.nasa.esdis.umm.processinglevelid"]/gmd:code/gco:CharacterString',
       source: {
         type: 'computed',
@@ -157,10 +176,6 @@ export const ISO_19115_SCHEME_EDITORS = {
     }
   }),
 
-  isotopiccategory: createIsoTopicCategoryEditor(),
-
-  productlevelid: createProductLevelIdEditor(),
-
   providers: createKeywordBlock('dataCentre', {
     fieldKeys: ['ShortName', 'LongName'],
     matchKeys: ['ShortName'],
@@ -170,16 +185,11 @@ export const ISO_19115_SCHEME_EDITORS = {
 
       return LongName ? `${ShortName} > ${LongName}` : ShortName
     }
-  })
+  }),
 
-  /*
-  Providers: short name not in examples
-              <gmd:organisationName>
-                <gco:CharacterString>National Snow and Ice Data Center</gco:CharacterString>
-              </gmd:organisationName>
-  rucontenttype: TBD
-  idnnode: no mapping
-  */
+  isotopiccategory: createIsoTopicCategoryEditor(),
+
+  productlevelid: createProductLevelIdEditor()
 }
 
 /**
--- a/serverless/src/shared/Iso19115DomEditor.js
+++ b/serverless/src/shared/Iso19115DomEditor.js
@@ -2,6 +2,17 @@ import Iso19115MetadataPathEditor from './Iso19115MetadataPathEditor'
 import { FULL_PATH_VALUE_FIELDS } from './redis-path-store/helpers/constants'
 
 /**
+ * A list of authorized domains used to validate the 'codeList' attribute
+ * within MD_KeywordTypeCode elements. This ensures only keywords originating
+ * from trusted metadata authorities are processed by the editor.
+ */
+const KEYWORD_DOMAINS = [
+  'cdn.earthdata.nasa.gov', // NASA Earthdata Resource Codelists
+  'www.isotc211.org', // ISO TC 211 Standard Codelists
+  'data.noaa.gov' // NOAA Metadata Authority Codelists
+]
+
+/**
  * Helper factory function to create a block editor configuration.
  * Maps a correction to an update operation within the editor instance.
  * @param {Object} config - Configuration object defining XPath and transformation logic.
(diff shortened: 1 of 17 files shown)


serverless/src/shared/Iso19115DomEditor.js
```diff
@@ -0,0 +1,199 @@
...
+    matchKeys: ['ShortName'],
+    getValue: ({ correction }) => {
+      const { ShortName } = correction.newKeywordObject
+      const LongName = correction.newLongName || ''
+
+      return LongName ? `${ShortName} > ${LongName}` : ShortName
+    }
+  }),
+
+  providers: createKeywordBlock('dataCentre', {
```
cgokey, line 332: The ISO19115 providers mapping looks pointed at the wrong XML structure. Right now [Iso19115DomEditor.js (line 175)](/Users/cgokey/Developer/nasa/kms/serverless/src/shared/Iso19115DomEditor.js#L175) treats providers as gmd:descriptiveKeywords/gmd:MD_Keywords with codeListValue='dataCentre', but the mapping sheet’s ISO19115 provider rows point to gmd:pointOfContact/gmd:CI_ResponsibleParty and the related distributor/processor/archive responsible-party paths instead. If a record follows the mapped responsible-party structure and does not also carry a dataCentre keyword block, this delegate will skip the provider update entirely for those places.

serverless/src/shared/Iso19115MetadataPathEditor.js
```diff
@@ -0,0 +1,235 @@
...
+
+  /**
+   * Updates complex keyword block nodes.
+   * Handles deletion of nested keywords and recursive cleanup of empty parents.
+   * @param {Object} correction - The change data.
+   * @param {Object} config - Configuration for the block node.
+   * @returns {boolean} True if the operation was successful.
+   */
+  updateBlockNode(correction, config) {
+    const targetNode = this.selectNodes(config.nodeXPath)[0] || null
```
cgokey: It seems like updateBlockNode() only searches the first matching MD_Keywords block because it does this.selectNodes(config.nodeXPath)[0].  If the target keyword lives in a later matching block, thinking we never look at it and the update/delete returns false even though the record contains a valid match. This probably needs to iterate across all matching blocks and stop on the first one that actually contains the keyword.

serverless/src/shared/Iso19115DomEditor.js
```diff
@@ -0,0 +1,199 @@
...
+        return acc
+      }, { Value: fullString.trim() })
+    }
+  },
+  replace: [
+    {
+      fieldPath: ({ node, editor }) => (editor.selectNodes('./gmx:Anchor', node).length > 0 ? 'gmx:Anchor' : 'gco:CharacterString'),
+      source: {
+        type: 'computed',
+        getValue: getValue || (({ correction }) => fieldKeys.map((k) => correction.newKeywordObject[k]).filter(Boolean).join(' > '))
```
cgokey: The full-path ISO serializer is collapsing empty hierarchy slots with .filter(Boolean).  The mapping html page says on ISO write that if an intermediate keyword is missing but a later keyword exists, we should fill the gap with NONE rather than removing the empty segment.

cgokey (approved):
Only a few comments, overall looks good.

serverless/src/shared/Iso19115DomEditor.js
htranho: Added additional paths.

serverless/src/shared/Iso19115MetadataPathEditor.js
htranho: Fixed.

serverless/src/shared/Iso19115DomEditor.js
htranho: Added 'NONE' for missing levels

serverless/src/shared/Iso19115MetadataPathEditor.js
```diff
@@ -0,0 +1,252 @@
...
+          const path = typeof replaceConfig.fieldPath === 'function'
+            ? replaceConfig.fieldPath({
+              node: matchingNode,
+              editor: this
+            })
+            : replaceConfig.fieldPath
+
+          const context = path.startsWith('//') ? this.document : matchingNode
+          const relativePath = path.startsWith('//') ? path : `./${path}`;
+          [fieldNode] = this.selectNodes(relativePath, context)
```
cgokey: Take a closer look at this, think it is only updating the first match here on this destructure.

serverless/src/shared/Iso19115MetadataPathEditor.js
```diff
@@ -0,0 +1,252 @@
...
+    const { matchingNode } = matchingData
+
+    // 2. Handle 'delete' action
+    if (correction.action === 'delete') {
+      const parentBlock = matchingNode.parentNode
+
+      // Clean up synchronized paths globally first
+      if (config.replace) {
+        config.replace
+          .filter((replConfig) => typeof replConfig.fieldPath === 'string' && replConfig.fieldPath.startsWith('//'))
```
cgokey, line 323: looks like delete iterates every node returned and removes them all. With the provider config’s //gmd:CI_ResponsibleParty/gmd:organisationName/gco:CharacterString, deleting one matched provider keyword would also remove unrelated responsible-party organization names elsewhere in the record. This seems too broad for provider cleanup and probably needs to be constrained to the mapped ISO target nodes?

serverless/src/shared/Iso19115DomEditor.js
```diff
@@ -0,0 +1,219 @@
...
+
+  locations: createKeywordBlock(
+    'place',
+    {
+      fieldKeys: FULL_PATH_VALUE_FIELDS.locations,
+      matchKeys: FULL_PATH_VALUE_FIELDS.locations
+    }
+  ),
+
+  platforms: createKeywordBlock('platform', {
```
cgokey, line 165: The platforms, instruments, and projects mappings currently only update the gmd:descriptiveKeywords/gmd:MD_Keywords text. The mapping workbook calls out additional ISO write targets for these values in the acquisition/identifier structures as well (for example platform rows, instrument row, and project rows). As written, a correction can update the visible keyword block while leaving the corresponding identifier/code/description fields stale, so I think these mappings are only partially implemented right now.

serverless/src/shared/Iso19115MetadataPathEditor.js
htranho: Fixed

serverless/src/shared/Iso19115MetadataPathEditor.js
htranho: Now constrained to the mapped ISO target nodes.

serverless/src/shared/Iso19115DomEditor.js
htranho: Now added paths to update the Acquisition nodes.

serverless/src/shared/Iso19115DomEditor.js
cgokey: So I think this still might be too broad, for example if the incoming provider correction is for ARCHIVER = NSIDC, this global path will update every CI_ResponsibleParty/organisationName with NSIDC, including unrelated roles like distributor. The workbook maps provider updates through specific role-based paths, so I think this should be constrained to the matching responsible-party locations for that provider role rather than every org-name occurrence in the document.

serverless/src/shared/Iso19115MetadataPathEditor.js
```diff
@@ -0,0 +1,377 @@
...
+      const oldShortName = correction.oldKeywordObject?.ShortName
+
+      // Only platforms and instruments have acquisition information sections
+      const isPlatformOrInstrument = correction.scheme === 'platforms' || correction.scheme === 'instruments'
+
+      // For platforms/instruments, delete from acquisition section too
+      if (isPlatformOrInstrument && oldShortName && config.replace) {
+        // Find all MD_Identifier nodes that match this platform/instrument
+        const allIdentifiers = this.selectNodes('//gmd:MD_Identifier', this.document)
+        const identifiersToDelete = allIdentifiers.filter((identifier) => {
```
cgokey, line 215: The acquisition cleanup here is matching identifiers by gmd:code text alone, but the ISO mapping is narrower than that. For platform/instrument sync we should also require the mapped gmd:codeSpace value, otherwise an unrelated acquisition identifier with the same short name can be updated or deleted too. Example: if one identifier is the UMM platform short-name field and another different identifier also happens to have code AQUA, this logic will treat both as the same target.

serverless/src/shared/Iso19115DomEditor.js
```diff
@@ -0,0 +1,344 @@
...
+            return correction.newLongName || ''
+          }
+
+          return node?.textContent || ''
+        }
+      }
+    ]
+  }),
+
+  projects: createKeywordBlock('project', {
```
cgokey, line 305: The project mapping has one more addition field to update. This code updates the project keyword block, but it does not also sync the mapped acquisition-side gmi:MI_Operation identifier fields from the workbook. Example: if MEASURES > Old Description becomes MEASURES-1 > New Description, the keyword block changes but MI_Operation can still be left with code MEASURES and the old description.

serverless/src/shared/Iso19115DomEditor.js
htranho: Since role (example ARCHIVER) is not part of correction object, we find it safer to not touch CI_ResponsibleParty.

serverless/src/shared/Iso19115MetadataPathEditor.js
htranho: Fixed.

serverless/src/shared/Iso19115DomEditor.js
htranho: Updated, added new paths

serverless/src/shared/Iso19115DomEditor.js
```diff
@@ -0,0 +1,360 @@
...
+    fieldKeys: ['ShortName', 'LongName'],
+    matchKeys: ['ShortName'],
+    getValue: ({ correction }) => {
+      const { ShortName } = correction.newKeywordObject
+      const LongName = correction.newLongName || ''
+
+      return LongName ? `${ShortName} > ${LongName}` : ShortName
+    },
+    // Sync with acquisition information section (gmi:MI_Operation)
+    additionalPaths: [
```
cgokey, line 315: This adds project sync into gmi:MI_Operation on replace, but I don’t see symmetric cleanup for delete in Iso19115MetadataPathEditor. That means we can remove the MD_Keywords project entry while leaving the synced acquisition operation values behind. Can we either add matching delete cleanup?

serverless/src/shared/Iso19115MetadataPathEditor.js
```diff
@@ -0,0 +1,407 @@
...
+    const targetNodes = this.selectNodes(config.nodeXPath)
+    if (!targetNodes || targetNodes.length === 0) return false
+
+    // Identify the first block that contains the matching node
+    const matchingData = targetNodes
+      .map((node) => ({
+        node,
+        matchingNode: this.findMatchingNode(node, correction, config)
+      }))
+      .find((data) => data.matchingNode !== null)
```
cgokey: Should this also guard undefined? findMatchingNode() appears to return undefined on a miss, so !== null may still pass.

serverless/src/shared/__tests__/applyIso19115MetadataCorrections.test.js
```diff
@@ -0,0 +1,1129 @@
...
+    expect(success).toBe(true)
+
+    const updatedXml = editor.serialize()
+
+    // Verify keyword block was updated
+    expect(updatedXml).toContain('MEASURES-1 &gt; New Project Description')
+    expect(updatedXml).not.toContain('MEASURES &gt; Making Earth System Data Records for Use in Research Environments')
+
+    // Verify acquisition MI_Operation code was updated
+    expect(updatedXml).toMatch(
```
eudoroolivares2016, line 775: Is there anyway that we can compare XML without having to add the spacing keys in the test assertions?

cgokey (changes requested):
I added a couple more comments

serverless/src/shared/Iso19115DomEditor.js
htranho: Added cleanup code

serverless/src/shared/Iso19115MetadataPathEditor.js
htranho: Fixed

serverless/src/shared/__tests__/applyIso19115MetadataCorrections.test.js
```diff
@@ -0,0 +1,1179 @@
...
+    const success = config(editor, correction)
+
+    expect(success).toBe(true)
+
+    const updatedXml = editor.serialize()
+
+    // Keyword should be removed
+    expect(updatedXml).not.toContain('MEASURES &gt; Making Earth System Data Records for Use in Research Environments')
+
+    // Note: The current implementation only handles updates via additionalPaths on replace actions.
```
cgokey, line 942: This looks outdated, the test looks like it handles it.

cgokey (approved):

eudoroolivares2016 (approved):

PR #116, opened 2026-07-07
KMS-690: Fix KMS authorizer handling of EDL level 5 bearer tokens created through the EDL "Generate Token" flow.

cgokey:
# Overview

### What is the feature?

Fix KMS authorizer handling of EDL level 5 bearer tokens created through the EDL "Generate Token" flow.

### What is the Solution?

This updates bearer-token validation in `fetchEdlProfile` so KMS no longer relies on `GET /oauth/userInfo` for bearer tokens. Instead, it:

- validates bearer/user tokens with `POST /oauth/tokens/user`
- authenticates that validation call with `Authorization: Basic <EDL_CLIENT_ID:EDL_PASSWORD>`
- decodes the JWT locally to read `assurance_level`
- keeps the existing Launchpad token flow unchanged
- adds `EDL_CLIENT_ID` to Bamboo/CDK/runtime wiring

- EDL authorizer bearer-token validation
- Bamboo/CDK runtime env wiring for `EDL_CLIENT_ID`
- Unit tests for `fetchEdlProfile` and authorizer behavior
- Deployment documentation in `README.md`

# Testing

1. Add `bamboo_EDL_CLIENT_ID` to Bamboo and deploy this branch. (done)
2. Call a protected KMS endpoint with an MMT-generated level 5 bearer token and verify access is still allowed.
3. Call the same protected KMS endpoint with an EDL-generated level 5 user token (`Bearer <token>`) and verify access is allowed.
4. Confirm the authorizer no longer fails on the EDL-generated token with `Unauthorized` / `403`.

I ran:

`KMS_AUTHORIZATION="Bearer ..." ALLOW_MUTATING_CORRECTION_ENDPOINT=true ./scripts/local/find_collection_with_positive_manual_sync_result.mjs  sit dif10 AMD_KOPRI`

which requires authentication and should work with both tokens (MMT generated one and EDL generated one)

# Checklist

## Summary by CodeRabbit

* **New Features**
  * Improved EDL bearer-token handling to validate tokens through the service’s token endpoint and return richer profile details.
  * Added support for a new required EDL client ID setting across deployment and runtime configuration.

* **Bug Fixes**
  * Updated profile lookup behavior and error handling for invalid tokens, missing configuration, and malformed validation responses.
  * Adjusted offline and Launchpad-related profile requests to use the correct EDL host in tests and deployment flows.


serverless/src/shared/fetchEdlProfile.js
```diff
@@ -81,31 +111,57 @@ const fetchProfileWithLaunchpadToken = async (host, launchpadToken) => {
...
+    throw new Error('EDL token validation response missing uid')
+  }
 
-  return buildProfile(profile)
+  const claims = decodeJwtClaims(edlToken)
+
+  return {
+    name: profile.uid,
+    uid: profile.uid,
+    assuranceLevel: claims.assurance_level
```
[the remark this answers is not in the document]
cgokey: /oauth/tokens/user is now the trust check for this flow. We only decode the bearer token after that call succeeds, and only to read assurance_level; token validity and uid come from the EDL-validated response, not from local JWT parsing.

htranho (approved):

serverless/src/shared/__tests__/fetchEdlProfile.test.js
```diff
@@ -5,12 +5,30 @@ import fetchEdlProfile from '../fetchEdlProfile'
 
 vi.mock('../fetchEdlClientToken', () => ({ default: vi.fn() }))
 
+const encodeBase64Url = (value) => Buffer
```
eudoroolivares2016, line 11: Would it just be easier to use a fake JWT token just generate one using https://www.jwt.io/ and save it to the repo we've done that in EDSC

serverless/src/shared/__tests__/fetchEdlProfile.test.js
```diff
@@ -20,73 +38,89 @@ afterEach(() => {
...
         expect(fetch).toHaveBeenCalledTimes(1)
-        expect(fetch).toHaveBeenCalledWith('https://sit.urs.earthdata.nasa.gov/oauth/userInfo', {
-          headers: {
-            Authorization: 'Bearer bearer-token-value',
-            'Content-Type': 'application/x-www-form-urlencoded;charset=UTF-8'
-          },
-          method: 'GET'
-        })
+        expect(fetch).toHaveBeenCalledWith(
+          'https://sit.urs.earthdata.nasa.gov/oauth/tokens/user',
```
eudoroolivares2016: is this defaulting to SIT if so we should mock the response where it fetches the env and use a fake one

eudoroolivares2016 (approved):

serverless/src/shared/__tests__/fetchEdlProfile.test.js
cgokey: I kept the JWT helper in the test rather than checking in a saved token fixture. We only need a minimal payload with assurance_level here, so generating it inline keeps the test self-contained and makes it obvious which claims matter.

serverless/src/shared/__tests__/fetchEdlProfile.test.js
cgokey: Good catch. I updated the test to mock getEdlConfig() and use an explicit fake EDL host, so it no longer relies on the default SIT config.

PR #118, opened 2026-07-15
KMS-694: Added POST for metadata correction handler

cgokey:
# Overview

### What is the feature?

Add a new asynchronous metadata correction endpoint for one or more collection concept ids. This keeps the existing sync endpoint unchanged while adding a fire-and-forget API that queues collection-scoped correction work through the existing FIFO processing flow.

### What is the Solution?

This change adds `POST /metadata_correction`, which accepts a JSON body with `collectionConceptIds`, validates the request, trims and deduplicates ids while preserving order, and publishes one metadata-correction request message per accepted collection concept id.

It also adds an optional `METADATA_CORRECTION_REQUEST_DELAY_MS` delay in the async consumer for requests originating from the new API. That delay is applied only to `source=metadataCorrectionApi` messages and uses the published `requestedAt` timestamp so we can reduce the chance of processing a collection while it is still being indexed after ingest.

Infrastructure updates thread the metadata-correction requests topic ARN into the KMS API Lambdas, add publish permissions for that topic, and expose the delay configuration to the processing stack. 

- KMS API Gateway/Lambda routing
- Async metadata correction request publishing
- Metadata correction consumer behavior
- CDK wiring for topic ARN, publish permissions, and delay configuration

# Testing

export KMS_AUTHORIZATION='Bearer <working level-5 token>'

curl --request POST \
  --header "Authorization: ${KMS_AUTHORIZATION}" \
  --header 'Content-Type: application/json' \
  --header 'Accept: application/json' \
  --data '{
    "collectionConceptIds": [
      "C1234567890-PROV", 
      "C1234567891-PROV"
    ]
  }' 
  "https://cmr.sit.earthdata.nasa.gov/kms/metadata_correction"

Tail cloudwatch logs in kms-sit-metadata-correction-service, kms-sit-request-metadata-correction

and check the audit log.

# Checklist

## Summary by CodeRabbit

* **New Features**
  * Added/extended the async metadata correction request endpoint to validate input, trim/deduplicate concept IDs, enqueue work, and return accepted/failed counts with details.
  * Introduced optional queued manual-request delay via `METADATA_CORRECTION_REQUEST_DELAY_MS`, configurable through stack settings (with clamping).
* **Bug Fixes**
  * Improved metadata correction service SQS batch handling to return `batchItemFailures` per failed record (by message ID) for correct partial retries.
* **Tests**
  * Expanded automated tests for request validation, partial publish outcomes, delay timing, and batch failure behavior.
  * Added local smoke tests covering end-to-end delay and partial batch failure scenarios.

### DIFF
--- a/serverless/src/metadataCorrectionService/handler.js
+++ b/serverless/src/metadataCorrectionService/handler.js
@@ -40,6 +40,8 @@ const buildBatchProcessingMetrics = ({
   return metrics
 }
 
+const MAX_METADATA_CORRECTION_REQUEST_DELAY_MS = 2000
+
 /**
  * Reads the optional async correction request delay from environment configuration.
  *
@@ -55,7 +57,7 @@ const getMetadataCorrectionRequestDelayMs = () => {
   const parsedValue = Number(rawValue)
 
   return Number.isFinite(parsedValue) && parsedValue > 0
-    ? parsedValue
+    ? Math.min(parsedValue, MAX_METADATA_CORRECTION_REQUEST_DELAY_MS)
     : 0
 }
 
--- a/serverless/src/metadataCorrectionService/handler.js
+++ b/serverless/src/metadataCorrectionService/handler.js
@@ -114,7 +114,8 @@ const delayQueuedManualRequestIfNeeded = async ({
  * flow can also be reused by synchronous API endpoints.
  *
  * @param {{ Records?: Array<{ body?: string, messageId?: string }> }} event - SQS batch event.
- * @returns {Promise<{batchItemFailures: Array}>} Empty batch failures for acknowledged messages.
+ * @returns {Promise<{batchItemFailures: Array<{itemIdentifier: string}>}>}
+ *   Failed message identifiers to retry through SQS partial batch response handling.
  */
 export const metadataCorrectionService = async (event) => {
   const records = event?.Records || []
@@ -165,16 +166,18 @@ export const metadataCorrectionService = async (event) => {
         recordCount: records.length
       }
     })
-
-    const firstRejectedResult = settledResults.find(({ status }) => status === 'rejected')
-
-    if (firstRejectedResult) {
-      throw firstRejectedResult.reason
-    }
   }
 
   return {
-    batchItemFailures: []
+    batchItemFailures: settledResults.reduce((failures, result, index) => {
+      const itemIdentifier = records[index]?.messageId
+
+      if (result.status === 'rejected' && itemIdentifier) {
+        failures.push({ itemIdentifier })
+      }
+
+      return failures
+    }, [])
   }
 }
 
--- a/serverless/src/metadataCorrectionService/handler.js
+++ b/serverless/src/metadataCorrectionService/handler.js
@@ -40,7 +40,7 @@ const buildBatchProcessingMetrics = ({
   return metrics
 }
 
-const MAX_METADATA_CORRECTION_REQUEST_DELAY_MS = 2000
+const MAX_METADATA_CORRECTION_REQUEST_DELAY_MS = 20_000
 
 /**
  * Reads the optional async correction request delay from environment configuration.
--- a/serverless/src/metadataCorrectionService/handler.js
+++ b/serverless/src/metadataCorrectionService/handler.js
@@ -1,3 +1,4 @@
+import { delay } from '@/shared/delay'
 import { CONSUMER_METRIC_NAMES } from '@/shared/emitConsumerMetrics'
 import { emitConsumerMetricsSafely } from '@/shared/emitConsumerMetricsSafely'
 import { logger } from '@/shared/logger'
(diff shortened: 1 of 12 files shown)


serverless/src/requestMetadataCorrection/handler.js
```diff
@@ -0,0 +1,159 @@
...
+          requestedAt
+        })
+
+        return {
+          collectionConceptId,
+          messageId: publishResult.messageId,
+          messageGroupId: publishResult.messageGroupId
+        }
+      })
+    )
```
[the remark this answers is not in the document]
cgokey: Addressed. I changed the batch publish path to Promise.allSettled so partial publish no longer collapses into a blind 500. The endpoint now returns 202 with separate accepted and failed lists when only part of the batch publishes, and still returns 500 when nothing was published.

serverless/src/metadataCorrectionService/handler.js
```diff
@@ -39,6 +40,70 @@ const buildBatchProcessingMetrics = ({
...
+  if (!rawValue) {
+    return 0
+  }
+
+  const parsedValue = Number(rawValue)
+
+  return Number.isFinite(parsedValue) && parsedValue > 0
+    ? parsedValue
+    : 0
+}
```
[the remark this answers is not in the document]
cgokey: Addressed. METADATA_CORRECTION_REQUEST_DELAY_MS is now clamped to 2000 in the consumer, so a bad config value can’t push the worker past its intended delay budget.

serverless/src/requestMetadataCorrection/handler.js
```diff
@@ -0,0 +1,159 @@
...
+  if (typeof body !== 'string' || body.trim().length === 0) {
+    throw new Error('Invalid metadata correction request: missing request body')
+  }
+
+  try {
+    return JSON.parse(body)
+  } catch {
+    throw new Error('Invalid metadata correction request: body must be valid JSON')
+  }
+}
```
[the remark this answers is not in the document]
cgokey: Addressed. The request parser now requires the body to deserialize into a JSON object, so literal null and array bodies return 400 instead of falling through into a 500. Added regression tests for both cases.

serverless/src/metadataCorrectionService/handler.js
htranho: Not related to this PR, but I see this array is never get loaded.

htranho: maybe instead of throwing error in 146:
batchItemFailures.push({ itemIdentifier: record.messageId })...

serverless/src/requestMetadataCorrection/handler.js
```diff
@@ -0,0 +1,215 @@
...
+  return 500
+}
+
+/**
+ * Converts a rejected publish reason into a stable response/log string.
+ *
+ * @param {unknown} error Rejection reason.
+ * @returns {string} Serialized error message.
+ */
+const toErrorMessage = (error) => {
```
eudoroolivares2016: Its only used once can we just do the error.ToString on the function or

serverless/src/metadataCorrectionService/handler.js
cgokey: Good catch. batchItemFailures is there because this Lambda uses the SQS batch response shape, but the current logic is still all-or-nothing: if any record fails we throw and retry the whole batch, otherwise we return an empty array. So you’re right that we are not currently using per-record batch failures here.

serverless/src/requestMetadataCorrection/handler.js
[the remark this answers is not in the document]
cgokey: This is addressed on the current branch: `requestMetadataCorrection/handler.js` it uses `Promise.allSettled`, returns `accepted`/`failed` results for partial publish success, and only returns `500` when nothing publishes. The partial-failure tests were added as well.

serverless/src/metadataCorrectionService/handler.js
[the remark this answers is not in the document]
cgokey: This is addressed as well in `metadataCorrectionService/handler.js` now clamps `METADATA_CORRECTION_REQUEST_DELAY_MS` to `20000ms`, and the test coverage was updated as well.

eudoroolivares2016 (approved):
Tested it against my running CMR REPL

htranho (approved):

serverless/src/requestMetadataCorrection/handler.js
cgokey: Done. I inlined the error stringification  and removed the helper since it was only used once.

serverless/src/metadataCorrectionService/handler.js
cgokey: Agree, as I looked at this some more.  I've updated this so the consumer now uses Lambda/SQS partial batch failure handling instead of failing the whole batch. I return failed messageIds in batchItemFailures, enabled reportBatchItemFailures on the SQS event source.

PR #121, opened 2026-07-24
KMS-699: Sanitize sparql input parameters

htranho:
# Overview

### What is the feature?

Sanitize user input parameters to used in sparql.

### What is the Solution?

Sanitize user input parameters to used in sparql.

All read and write interfaces with parameters

# Testing

All interfaces with input parameters.

### Attachments

N/A

# Checklist

## Summary by CodeRabbit

- **Security Improvements**
  - Strengthened sanitization and escaping for SPARQL query/update construction across concept, scheme, labels, relation change notes, and date/version inputs.
  - Added more robust handling of encoded and control/special characters to better prevent malformed or unsafe SPARQL.
  - Added clearer input validation with explicit error handling for invalid values.

- **Bug Fixes**
  - Improved resilience to malformed or improperly encoded inputs that previously could lead to incorrect query behavior.

- **Tests**
  - Expanded automated test coverage for escaping/sanitization edge cases and safer query generation.


cgokey:
Take a look at addChangeNotes.js around line 57. The change-note SPARQL update is still built from raw relation values (for example fromPrefLabel, toPrefLabel, and from) without going through the new taint-checking/escaping path, so this looks like a remaining write-side SPARQL injection surface even though the read/query builders were tightened in this PR.

serverless/src/shared/operations/queries/getConceptsQuery.js
```diff
@@ -1,17 +1,23 @@
 import prefixes from '@/shared/constants/prefixes'
+import { escapeSparqlString } from '@/shared/escapeSparqlString'
+import { sanitizeScheme } from '@/shared/sanitizeScheme'
 
-export const getConceptsQuery = (conceptScheme, pattern, limit = 1000, offset = 0) => `
+export const getConceptsQuery = (conceptScheme, pattern, limit = 1000, offset = 0) => {
+  const safeConceptScheme = sanitizeScheme(conceptScheme)
```
cgokey, line 6: One thing I think we should still tighten here. Right now conceptScheme is sanitized, but if a caller sends an invalid value that sanitizes down to an empty string, we silently drop the scheme filter and broaden the query instead of rejecting the input. Could we mirror the validation pattern already used in getTotalCountQuery and fail here too?

```suggestion
let safeConceptScheme = ''
if (conceptScheme) {
  safeConceptScheme = sanitizeScheme(conceptScheme)
  if (!safeConceptScheme) {
    throw new Error('Invalid conceptScheme provided')
  }
}
```

cgokey: Also take a look at getCreateRelationshipQuery.js:6-8 and getDeleteRelationshipQuery.js:6-8. sourceUuid and each targetUuid are still interpolated directly into concept IRIs here, so these helpers are not yet using the same taint-checking pattern the rest of KMS-699 added. Could we validate the UUIDs first (for example with sanitizeConceptId or a shared toConceptIRI helper) before building these triples?

cgokey: Also take look at getMetadataCorrectionAuditLog.js:130-135 still builds FILTER clauses with metadataCorrectionAudit.js:17-21, which is a separate partial escaping helper and doesn’t use the same sanitizer/escaping behavior as the rest of this PR. Could we switch this path over to the shared SPARQL escaping logic?

cgokey: One broader concern I still have with the sanitizer approach is that, for structured inputs, it rewrites invalid values instead of preserving whether they were actually valid. That means a bad non-empty input can collapse into '' or into a different identifier, so downstream code can no longer distinguish "filter omitted" from "filter supplied but invalid," and in some paths that can fail open or point at the wrong resource. I think the safer pattern for structured values like scheme/conceptId/IRIs is: valid input stays unchanged, missing input stays missing, and invalid non-empty input comes back as a distinct invalid result rather than being normalized into something else.
```
export const sanitizeScheme = (scheme) => {
  if (scheme == null || scheme === '') {
    return ''
  }

  if (typeof scheme !== 'string') {
    return null
  }

  const sanitized = scheme.replace(schemeRegex, '')

  return sanitized === scheme ? sanitized : null
}
```
Then callers do this:

```
const safeScheme = sanitizeScheme(scheme)

if (scheme && safeScheme === null) {
  throw new Error('Invalid scheme provided')
}
```

htranho:
> Take a look at addChangeNotes.js around line 57. The change-note SPARQL update is still built from raw relation values (for example fromPrefLabel, toPrefLabel, and from) without going through the new taint-checking/escaping path, so this looks like a remaining write-side SPARQL injection surface even though the read/query builders were tightened in this PR.

Fixed

serverless/src/shared/operations/queries/getConceptsQuery.js
htranho: > One thing I think we should still tighten here. Right now conceptScheme is sanitized, but if a caller sends an invalid value that sanitizes down to an empty string, we silently drop the scheme filter and broaden the query instead of rejecting the input. Could we mirror the validation pattern already used in getTotalCountQuery and fail here too?

Fixed

htranho: > One broader concern I still have with the sanitizer approach is that, for structured inputs, it rewrites invalid values instead of preserving whether they were actually valid. That means a bad non-empty input can collapse into '' or into a different identifier, so downstream code can no longer distinguish "filter omitted" from "filter supplied but invalid," and in some paths that can fail open or point at the wrong resource. I think the safer pattern for structured values like scheme/conceptId/IRIs is: valid input stays unchanged, missing input stays missing, and invalid non-empty input comes back as a distinct invalid result rather than being normalized into something else.
> 
> ```
> export const sanitizeScheme = (scheme) => {
>   if (scheme == null || scheme === '') {
>     return ''
>   }
> 
>   if (typeof scheme !== 'string') {
>     return null
>   }
> 
>   const sanitized = scheme.replace(schemeRegex, '')
> 
>   return sanitized === scheme ? sanitized : null
> }
> ```
> 
> Then callers do this:
> 
> ```
> const safeScheme = sanitizeScheme(scheme)
> 
> if (scheme && safeScheme === null) {
>   throw new Error('Invalid scheme provided')
> }
> ```

Fixed

htranho: > Also take a look at getCreateRelationshipQuery.js:6-8 and getDeleteRelationshipQuery.js:6-8. sourceUuid and each targetUuid are still interpolated directly into concept IRIs here, so these helpers are not yet using the same taint-checking pattern the rest of KMS-699 added. Could we validate the UUIDs first (for example with sanitizeConceptId or a shared toConceptIRI helper) before building these triples?

Fixed

htranho: > Also take look at getMetadataCorrectionAuditLog.js:130-135 still builds FILTER clauses with metadataCorrectionAudit.js:17-21, which is a separate partial escaping helper and doesn’t use the same sanitizer/escaping behavior as the rest of this PR. Could we switch this path over to the shared SPARQL escaping logic?

Fixed

serverless/src/shared/operations/queries/getCreateRelationshipQuery.js
```diff
@@ -1,11 +1,25 @@
...
+    if (safeUuid === null) {
+      throw new Error('Invalid targetUuid provided')
+    }
+
+    return safeUuid
+  })
+  const createDirectTriples = () => safeTargetUuids.map((uuid) => `<https://gcmd.earthdata.nasa.gov/kms/concept/${safeSourceUuid}> ${relationship} <https://gcmd.earthdata.nasa.gov/kms/concept/${uuid}> .`).join('\n    ')
 
-  const createInverseTriples = () => targetUuids.map((uuid) => `<https://gcmd.earthdata.nasa.gov/kms/concept/${uuid}> ${inverseRelationship} <https://gcmd.earthdata.nasa.gov/kms/concept/${sourceUuid}> .`).join('\n    ')
+  const createInverseTriples = () => safeTargetUuids.map((uuid) => `<https://gcmd.earthdata.nasa.gov/kms/concept/${uuid}> ${inverseRelationship} <https://gcmd.earthdata.nasa.gov/kms/concept/${safeSourceUuid}> .`).join('\n    ')
```
eudoroolivares2016, line 22:
```suggestion
  const createInverseTriples = () => safeTargetUuids.map((uuid) => `<https://gcmd.earthdata.nasa.gov/kms/concept/${safeSourceUuid}> ${inverseRelationship} <https://gcmd.earthdata.nasa.gov/kms/concept/${safeSourceUuid}> .`).join('\n    ')
```

Do these need to be safe source?

serverless/src/shared/updateVersionMetadata.js
```diff
@@ -46,30 +47,56 @@ export const updateVersionMetadata = async ({
...
+    : undefined
+  if (lastSynced !== undefined && safeLastSynced === null) {
+    throw new Error('Invalid lastSynced provided')
+  }
+
   let deleteClause = ''
   let insertClause = ''
 
   if (version !== undefined) {
     deleteClause += `<${versionUri}> gcmd:versionName ?oldVersionName .\n`
```
eudoroolivares2016, line 83:
```suggestion
    deleteClause += `<${safeVersion}> gcmd:versionName ?oldVersionName .\n`
```

do these also have to be the safe versions?

eudoroolivares2016 (approved):

serverless/src/shared/captureRelations.js
```diff
@@ -39,6 +41,8 @@ export const captureRelations = async (conceptId, version, transactionUrl = null
     return uri.split('#').pop()
   }
 
+  const safeConceptId = sanitizeConceptId(conceptId)
```
cgokey, line 44: Looks like we are sanitizing conceptId here, but we also need to check whether safeConceptId is null before building the query. Otherwise an invalid conceptId can fall through and produce something like https://gcmd.earthdata.nasa.gov/kms/concept/null instead of failing closed.

cgokey (approved):

serverless/src/shared/operations/queries/getCreateRelationshipQuery.js
htranho: yes, the uuid can be invalid

serverless/src/shared/updateVersionMetadata.js
htranho: those can be invalid/unsafe

